PR #6, opened 2020-05-07
rename to "Asset Compute Worker SDK" and other readme changes

alexkli:
"Asset Compute SDK" is a bit confusing as it does not include a way to call the service, for which we have the separate [asset-compute-client](https://github.com/adobe/asset-compute-client) library. And we need to include "Adobe" in the name. Hence I renamed it to "Adobe Asset Compute Worker SDK".

(The npm & repo name `asset-compute-sdk` is fine as is IMO).

Also improved text a bit & fixed some formatting. And removed mentioning of New Relic (shouldn't be in here IMO).

And added npm, license and travis badges.


alexkli:

PR #9, opened 2020-05-28
NUI-552 [Travis] All Action Code Open Source Repos Should Only be Tested on Linux

DrGirlfriend:
[NUI-552 [Travis] All Action Code Open Source Repos Should Only be Tested on Linux
](https://jira.corp.adobe.com/browse/NUI-552)

Removed Node versions 10.15 and 10.19 from Travis
Modified Travis to run only on Linux
Added code coverage reporting
Simplified script test
Added script beautify
Updated eslint version and fixed complaints


.travis.yml
```diff
@@ -9,19 +9,18 @@
...
 
 language: node_js
-os:
-  - linux
-  - osx
 
 node_js:
   - "10"
-  - "10.15"
-  - "10.19"
```
alexkli, line 19: `10.15` and `10.19` are needed for actions.

- cli related repos:
  - os: linux, windows & mac
  - node: 10, 12, 14
- nui service code repos:
  - os: linux
  - node: 10, 12, 14 plus whatever versions we use in prod, 10.15, 10.19 right now
- used in both:
  - superset of the above

alexkli:

PR #11, opened 2020-06-06
NUI-600 Option to inline small renditions in IO events

alexkli:
- add `inlineLimit` option to renditions
  - if set on rendition instructions AND rendition size is <= inlineLimit
    - will add content into new `data` field on rendition event, encoded as data-uri
  - if set, but too large, will upload as usual
  - if not set, will always upload (opt-in behavior)
  - maximum allowed value is 32kb, if exceeded, will always upload
  - note: AEM would likely use 16kb, IO Event payload is capped ~64kb
- improving Rendition object abstraction
  - since mime type handling for rendition had to be reused
  - encapsulated reading of metadata inside Rendition
  - existence check done with Rendition.exists()
- added placeholder for dc:format and encoding fields in renditions
  - but not enabled yet, tbd in separate PR
- new env var ASSET_COMPUTE_SDK_DISABLE_CGROUP_METRICS to disable cgroup metrics
  - intended for unit tests

Example snippet sending it:
```json
  "renditions": [{
    "fmt": "txt",
    "inlineLimit": 16384,
    "target": "https://blobstore.com/container/rendition"
  }]
```

Example IO event with rendition data inlined:
```json
{
  "type": "rendition_created",
  "date": "2020-06-06T06:40:51.304Z",
  "data": "data:text/plain;charset=utf-8;base64,cmVuZGl0aW9uIGNvbnRlbnQgc21hbGxlciB0aGFuIDE2a2I=",
  "metadata": {
    "repo:size": 35,
    "repo:sha1": "c90e676b979ff146c13c22ef653f5ecf0a71d5bb"
  },
  "rendition": {
    "fmt": "txt",
    "inlineLimit": 16384
  }
}
```


alexkli:
Messed up commit history, creating a new PR.

PR #12, opened 2020-06-06
[Prototype] NUI-600 Option to inline small renditions in IO events

alexkli:
For [NUI-600](https://jira.corp.adobe.com/browse/NUI-600).

**Note this is a prototype for now.** Intended for measuring the performance impact first.

- add `inlineLimit` option to renditions
  - if set on rendition instructions AND rendition size is <= inlineLimit
    - will add content into new `data` field on rendition event, encoded as data-uri
  - if set, but too large, will upload as usual
  - if not set, will always upload (opt-in behavior)
  - maximum allowed value is 32kb, if exceeded, will always upload
  - note: AEM would likely use ~16kb~ 100 bytes, IO Event payload is capped ~64kb
- improving Rendition object abstraction
  - since mime type handling for rendition had to be reused
  - encapsulated reading of metadata inside Rendition
  - existence check done with Rendition.exists()
- added placeholder for dc:format and encoding fields in renditions
  - but not enabled yet, tbd in separate PR
- new env var ASSET_COMPUTE_SDK_DISABLE_CGROUP_METRICS to disable cgroup metrics
  - intended for unit tests

Example snippet sending it:
```json
  "renditions": [{
    "fmt": "txt",
    "inlineLimit": 16384,
    "target": "https://blobstore.com/container/rendition"
  }]
```

Example IO event with rendition data inlined:
```json
{
  "type": "rendition_created",
  "date": "2020-06-06T06:40:51.304Z",
  "data": "data:text/plain;charset=utf-8;base64,cmVuZGl0aW9uIGNvbnRlbnQgc21hbGxlciB0aGFuIDE2a2I=",
  "metadata": {
    "repo:size": 35,
    "repo:sha1": "c90e676b979ff146c13c22ef653f5ecf0a71d5bb"
  },
  "rendition": {
    "fmt": "txt",
    "inlineLimit": 16384
  }
}
```


lib/rendition.js
```diff
@@ -41,12 +74,74 @@ class Rendition {
...
+
     size() {
-        return this.metadata[RenditionMetadata.REPO_SIZE];
+        return fs.statSync(this.path).size;
     }
 
-    sha1() {
-        return this.metadata[RenditionMetadata.REPO_SHA1];
+    async sha1() {
+        if (!this._sha1) {
```
tmathern, line 86: Nitpick - but I would add a check verifying the `this._sha1` is not blank (since nothing is really private in javascript, something could have changed it to a blank string).

By blank I mean containing chars like `\n` or `\t` only.

lib/rendition.js
```diff
@@ -60,6 +155,26 @@ class Rendition {
         return obj;
     }
 
+    inline() {
+        return Number.isInteger(this.instructions.inlineLimit)
```
tmathern, line 159: So, no parsing if `this.instructions.inlineLimit` is a string?

lib/rendition.js
```diff
@@ -60,6 +155,26 @@ class Rendition {
...
+            && this.instructions.inlineLimit <= INLINE_LIMIT_MAX
+            && this.size() <= this.instructions.inlineLimit;
+    }
+
+    asDataUri() {
+        // should never happen as inlineLimit is limited, but extra safeguard against reading large files
+        if (this.size() > DATA_URI_LIMIT) {
+            throw new GenericError(`Rendition too large for data uri ${this.name}`);
```
tmathern, line 167: We have a `RenditionTooLarge` error. Would it be worthwhile to use it and have this as clarifying error message?

lib/rendition.js
```diff
@@ -60,6 +155,26 @@ class Rendition {
...
+        }
+        const data = fs.readFileSync(this.path);
+
+        // cannot have spaces in data uri
+        // convert 'text/plain; charset=utf-8' into 'text/plain;charset=utf-8'
+        const type = this.contentType().replace(" ", "");
+
+        return `data:${type};base64,${data.toString("base64")}`;
```
tmathern, line 175: Do we want to add a safety layer verifying the data uri is valid before sending? E.g. using https://www.npmjs.com/package/valid-data-url.

lib/rendition.js
alexkli: Why? This is an internal variable (`_`) to cache the value of the potentially very expensive sha calculation of large files, and it needs to only run if not set yet.

If a code change would accidentally modify it, then it should not be hidden and lead to another run.... and such a modification could always unset it as well (e.g. `_sha1 = false`), which we cannot protect from in the condition here.

lib/rendition.js
alexkli: Yes, it has to be a number in the request json. No need to allow strings IMO, just increases the complexity.

lib/rendition.js
alexkli: No, as this would not be a _client_ error (which `RenditionTooLarge` is) but an error in our code, which should be visible as such.

With the current code it should be technically impossible to happen because of the other [check here](https://github.com/adobe/asset-compute-sdk/pull/12/files/6b89dea5624f7485801e9b75cd89404cb62e5710#diff-7758ccf59648f08e27a97a2510f1c2dfR160) as the comment says. It is just an extra safe guard in case the logic gets messed up in the future, since the file is read below, and we don't want to read large files by mistake.

lib/rendition.js
alexkli: We are building it according to the spec here...

lib/worker.js
```diff
@@ -129,22 +127,24 @@ class AssetComputeWorker {
...
-            delete metrics_object.cpuacct.stat;
-            if (this.previousCpuUsage) {
-                metrics_object.cpuacct.usagePercentage = cpu.calculateUsage(this.previousCpuUsage, curr_cpu_usage);
-            } else {
-                metrics_object.cpuacct.usagePercentage = undefined;
-            }
-            this.previousCpuUsage = curr_cpu_usage;
-            return metrics_object;
-        });
+        if (!process.env.ASSET_COMPUTE_SDK_DISABLE_CGROUP_METRICS) {
```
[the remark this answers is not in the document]
tmathern: I think the name would be too long then. Also, I could totally understand if people use the var to deactivate cgroup metrics when not in test: I could see this coming in custom workers. Is it a hack? Yes. Can we prevent it? Not really.

lib/rendition.js
tmathern: Validate for inline() ? What do you mean?

lib/rendition.js
```diff
@@ -41,12 +74,74 @@ class Rendition {
         }
     }
 
+    exists() {
+        return fs.existsSync(this.path);
+    }
+
```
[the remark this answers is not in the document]
alexkli: Hmm, I wouldn't like that. We could debate using `fs.promises` instead and `await rendition.exists()` accordingly...

lib/worker.js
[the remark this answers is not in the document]
alexkli: Ack. I think for these env vars we recently went with a specific name that has a prefix based on the module, i.e. `ASSET_COMPUTE_SDK_*`. That makes them unique. There could be other use cases to disable it than just tests.

lib/rendition.js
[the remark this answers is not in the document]
tmathern: I would keep it simple for now - and sync, so the changeset is minimal. If we "asyncify" it, there would be more changes in the SDK, right?

Let's first validate the prototype, get some data about performance, then we can start tweaking. Wdyt?

lib/rendition.js
pheenomenon: the calling client does it [here](https://github.com/adobe/asset-compute-sdk/blob/6b89dea5624f7485801e9b75cd89404cb62e5710/lib/worker.js#L277) , so it seems like the logic can be encapsulated inside asDataUri()

tmathern: I think I like the way it is currently separated best. `asDataUri()` and `inline()` doing the most minimal job. Also the condition you point to is semantically clearer for a reader to understand... ? Or is it just my biased opinion?

lib/rendition.js
```diff
@@ -41,12 +74,74 @@ class Rendition {
...
+    charset() {
+        // TODO: charset should be set by worker or determined by actually looking at the file
+        return mime.charset(path.extname(this.path));
+    }
+
+    contentType() {
+        // TODO: mimetype should be set by worker (this only looks at extension)
+        return mime.contentType(path.extname(this.path)) || DEFAULT_CONTENT_TYPE;
```
alexkli, line 104: That we  "guess" the mime type based on the filename extension here (this is all what `mime.contentType()` does) is actually a problem already as AEM still requests text renditions using the old `fmt=text-main`, so the rendition files are getting named `rendition0.text-main` and then it sees it as `application/octet-stream` since the mime library has no idea what `text-main` is.

As the comment says we need a way for workers to reliably communicate the actual mime type to the SDK for use here. The workers are the ones that know and it should not depend on that temporary rendition filename that the SDK comes up with.

This needs a separate channel, some options:
- static: set in `worker()` callback through some API, i.e. saying "this worker always generates mime type X"
- dynamic and/or shell script workers: store in a side car file (similar to `errors.json`)
- ...

lib/rendition.js
pheenomenon: yes, better readability this way, just wonder if these(`asDataUri() and inline()`) functionally always go together.

lib/rendition.js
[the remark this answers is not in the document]
alexkli: ack

lib/rendition.js
tmathern: I would tend to go towards and API rather than using sidecar-files:
- files can "disappear" or fail to be written;
- I think it would be easier to reuse and extend the logic with an API/ callback in the SDK. So, if it is set by the worker, the worker, should always know which kind of file it generated... Otherwise how would it generate the proper file? The worker would be best placed to set the contenttype, then. Also, long long time ago, I started a toy library `node-file-mimetypes` (not public) to do that contentype recognition based on the file. I'm pretty sure we could adapt it to be used here. (The reason why we don't use it yet is the repo's test folder size. I have ideas to reduce it, but never implemented it)

alexkli: Yeah, but shell script workers have no API to call, exchanging a file is the only good mechanism. Environment variables don't work in the direction of child process to parent.

alexkli: created https://jira.corp.adobe.com/browse/NUI-607

pheenomenon:
> 1. inlineLimit as opt-in feature ok?
Agree. Unless we want to default to this behavior, in which case clients would have to change code at their end(I think). 

> 3. maybe inlineLimit should be inside the rendition.target object instead?
So we want our client to pass us rendition target and `inlineLimit`, but they do_not_want/cannot determine the rendition size. iiuc, when `inlineLimit` is passed in then we don't care about `target`. So, cant see when "inlineLimit" will have relevance outside "target" context. so something like ? 
 "renditions": [{
    "fmt": "txt",
    "target": {
            "destination": "https://blobstore.com/container/rendition"
              "inlineLimit":  16384
      }
  }]

> 4.  name data property in event?
+1 , how about `inLineRenditionData` ?

lib/storage/http.js
```diff
@@ -51,13 +42,12 @@ async function upload(rendition) {
...
 
     try {
-        const contentType = mime.contentType(path.extname(file)) || DEFAULT_CONTENT_TYPE;
         if (typeof target === 'string') {
             console.log(`uploading rendition ${file} to ${AssetComputeLogUtils.redactUrl(target)}, size = ${rendition.size()}`);
             await http.uploadFile(file, target, {
                 retryEnabled: !process.env.ASSET_COMPUTE_DISABLE_RETRIES,
                 headers: {
-                    'content-type': contentType
+                    'content-type': rendition.mimeType()
```
alexkli, line 50: This should use `rendition.contentType()` instead to include `charset` in case of text files), which was done by `mime.contentType()` used before.

alexkli:
FYI, this will have to be rebased on top of #14.

alexkli:
Productized in #30.

PR #14, opened 2020-06-12
FEATURE-RELEASE: Set mimetype and encoding on renditions (using generated rendition file)

tmathern:
## Description
- Adds content type (aka mime type) recognition for a rendition.
- Adds encoding recognition for a rendition.
- Both encoding and content type are metadata. Metadata are now filled out during the `rendition_success` call, and sent in the event. This should cover shellscript and javascript workers.

- Add a `unit-tests` script to the package.json, which displays log statements in the console (useful for debugging) and generates a code coverage version in HTML (helps to find locations where code coverage can be improved).

**_Technical details:_**
To add mimetype and encoding recognition, I ended up calling the `file` command from the underlying operating system. This relies on libmagic.

Libmagic has bindings for nodejs, but adding any of those libraries would make action packaging fail:
- for [mmmagic](https://www.npmjs.com/package/mmmagic): this library tries to recompile its native dependency during installation (npm-i). There are ways to compile for different target than we install on, so before packaging we could recompile for a Linux target... This library will work on Mac and Linux, but make our packaging fail with a path resolution error. Actually, webpack can't/won't support the packaging of native libraries/binaries (*.node, *.o for instance). Since we would not be able to package, and I didn't want every single worker to have to either know where libmagic/a magic file is found on a system or use a custom Docker image we would maintain, I switched to another solution.
- for [magic](https://www.npmjs.com/package/@npcz/magic): this library uses the native implementation of libmagic, written in C and transpiled to web assembly. I had similar issues for packaging as for the above library with this, and also I could not get it to run on my machine properly (web assembly kept complaining it can't initialize). 
- other libraries relying on libmagic are either old or unmaintained.
- [file-type](https://www.npmjs.com/package/file-type) library is full javascript, but is unfortunately incomplete and cannot recognize everything (allthough for the time being it would cover our needs).

So I ended up using `execa` to call the `file` command from the underlying operating system. I chose [execa](https://www.npmjs.com/package/execa) because it is an improved `child_process` (e.g. cleans up if parent process dies), that makes it easy to read results from the command. Also, it is promisified and runs cross-platform (if Windows had a `file` installed...). 

To finish, I also did some sha1 handling changes and moved metata generation into the rendition class.

Fixes # [NUI-607](https://jira.corp.adobe.com/browse/NUI-607)

## Types of changes

## Checklist:

- [x] I have added tests to cover my changes (and some more to increase coverage a bit).


tmathern:
Based on prototype from https://github.com/adobe/asset-compute-sdk/pull/12.

lib/rendition.js
```diff
@@ -42,17 +73,109 @@ class Rendition {
...
+    async mimeType() {
+        let mimetype = DEFAULT_CONTENT_TYPE;
+
+        // Long version of the command has the advantage of ensuring same results on Linux and Mac
+        if(this.exists() && this.size() > 0){
+            // mimetype = DEFAULT_CONTENT_TYPE; // if the file can't be red, should we return undefined?
+            const commandResult = await execa('file', ['-b', '--mime-type', this.path]);
+            if(commandResult.exitCode === 0){
+                mimetype = commandResult.stdout;
+            }
```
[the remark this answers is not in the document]
tmathern: `execa` removed the need for the trim, and adds the exit code.
Also, for some reason I prefer libraries which already work as promises and do not need the util-promisification.
But we can live without it.

lib/rendition.js
```diff
@@ -42,17 +73,109 @@ class Rendition {
     }
 
     size() {
-        return this.metadata[RenditionMetadata.REPO_SIZE];
+        return fs.statSync(this.path).size;
```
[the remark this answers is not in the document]
tmathern: I don't think so. I left some of then voluntarily sync, like `exist` and `size` so they may be used in non-async location.
If we promisify everything, I would suggest doing it later.

test/rendition.content.test.js
```diff
@@ -0,0 +1,135 @@
...
+        // overwrite path to point to test files
+        rendition.path = './test/files/file.bmp';
+        let result = await rendition.mimeType();
+        assert.strictEqual(result, 'image/x-ms-bmp');
+
+        rendition.path = './test/files/file.tif';
+        result = await rendition.mimeType();
+        assert.strictEqual(result, 'image/tiff');
+
+        rendition.path = './test/files/file with spaces in name.bmp';
```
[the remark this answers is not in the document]
tmathern: After using library which install near 700MB of dependency during their initialization, I will not comment further on this :D

package.json
```diff
@@ -60,6 +63,6 @@
     "test": "nyc mocha test --file test/logfile.setup.js --recursive --exit",
     "posttest": "eslint ./ && license-checker --summary && ./node_modules/lockfile-lint/bin/lockfile-lint.js --path package-lock.json —type npm --allowed-hosts npm artifactory.corp.adobe.com --allowed-schemes \"https:\" \"file:\" --empty-hostname true",
     "beautify": "eslint ./ --fix",
-    "unit-test": "nyc mocha test --file test/logfile.setup.js --recursive --exit"
+    "unit-test": "nyc --reporter=html --reporter=text mocha test --recursive --exit -- -v"
```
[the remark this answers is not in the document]
tmathern: I prefer the html report, personally. I find it more readable to track down uncovered parts.

package.json
```diff
@@ -19,12 +19,15 @@
     "ajv": "6.12.2",
     "clone": "^2.1.2",
     "data-uri-to-buffer": "^3.0.1",
+    "deprecate": "^1.1.1",
```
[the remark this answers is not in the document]
tmathern: It's not used - it's leftover from when I was trying some things out.

tmathern:
@alexkli 
`Also the rendition.exists() usage in worker.js (at least 2).` is not part of mimetype feature (it's more a refactoring), and could/should be another PR. It was not clear to me I had to productize all NUI-600 prototype.

tmathern:
@alexkli 

Regarding content detection feature:
I was pondering, when coding it, why we need a mechanism to set mimetype/encoding type from a worker, and settled with myself to only code automated detection for now (for me it dud fit the bill to have a mimetype-setting mechanism, as it is done in the sdk at rendition-success event sending time. Also, I could share implementation for shell and js workers).
I can see why we may one day need it for less common types. But currently I would have a tendency to rely on mime identification rather than setting:
- if sdk does the identification in one place, a developer doesn't need to remember to do it (which would be the case for setting a mime type manually, right? I mean it would have to be done in the worker, not the sdk)
- using detection based on our container should always yield the same mime type (there are different mimetype fitting the same file type, eg image/tif and image/tiff for tif files. There are even more synonyms for psd or ms office files).

I will add the mechanism for setting a mimetype from a worker as this is what you wish for. But I will be keeping automated detection as fallback, or even only doing automated detection in the sdk, be it only for the reasons I listed above.

alexkli:
@tmathern 
> Also the rendition.exists() usage in worker.js (at least 2). (tmn edit: changed - but I could find only one. The other one may be part of the inline() feature)

There are 2:
- https://github.com/adobe/asset-compute-sdk/pull/12/files#diff-644d8ede25bb88427ccf4229d03bba1bR179
- https://github.com/adobe/asset-compute-sdk/pull/12/files#diff-644d8ede25bb88427ccf4229d03bba1bR239

tmathern:
I added a mechanism so workers can set the mimetype and encoding of the rendition they're creating. I set mimetype and encoding together (1 function call), as they are related. Also, I set them together because I updated them together when using the file command too.

**Usage:**
- for shellscript workers: the worker should write mimetype+encoding in a file in the `/out` directory of the worker (`mimeinfo-${renditionId}.txt`). The exact path is made available in an ENV variable, `mimefile`.The content should be formatted like the `file` command would (aka `mimetype; encoding`). The file has the rendition id in its name to ensure it is unique and can be recognized as a given rendition's mime information. The file is created by the sdk the following way during the run preparation (prepare.js): 
```
function prepareMimeInfoFile(directory, renditionId) {
    const mimeDir = path.resolve(directory, "mimeinfo");
    fs.mkdirSync(mimeDir, {recursive: true});
    return path.resolve(mimeDir, `mimeinfo-${renditionId}.txt`);
}
```
- for javascript workers: javascript workers can call the method `setMimeInformation(mime, encoding)` on the rendition object.

**Caching:**
- `setMimeInformation` is used to update the mime+encoding information for the rendition. If `setMimeInformation` gets a mimetype and encoding entered as parameters, those have precedence. If not, `setMimeInformation` tries to read the mime-info file. If this file is not found, it falls back on the `file` command (if available). If the `file` command is not here (which is the case currently in our images), it tries to find the mimetype using the file-type library (https://www.npmjs.com/package/file-type).
- I have been debating with myself whether or not to cache mime and encoding information (to avoid the possible costly `file` fallback). In the end, I decided not to cache yet. This will allow information to be re-read at each `.encoding()` or `.mimeType()` call, which means that a mimetype can be updated (may be useful to update a mimetype as more information becomes available). Wdyt?

**Known issues:**
- some mimetype can have synonyms. For instance, mimetypes `image/tif` and `image/tiff` are synonyms. Since we cannot really control what the final mimetype name will be, we should not do tests based on that. The fix would be to normalize names (using a map, for instance). Do we need this? I don't think so for now (but maybe later).

**Requests for feedback:**
- Wdyt about the usage to set mimetypes?
- For me, mimetype and encoding are deeply linked, hence handled together. But would you prefer them to be split: one method for setting encoding, one for mimetype?
- Content-type is generated based on the content of mimetype and encoding properties. Are there objections to that?

**Left to do:**
- If usage is OK, I'll add more unit tests.

lib/rendition.js
```diff
@@ -12,28 +12,59 @@
...
 const RENDITION_BASENAME = 'rendition';
+const DEFAULT_CONTENT_TYPE = 'application/octet-stream';
+
+const HASH_ALGORITHM = 'sha1';
+
+const Metadata = {
+    REPO_SIZE: 'repo:size',
+    REPO_SHA1: 'repo:sha1',
+    DC_FORMAT: 'dc:format',
+    ENCODING: 'repo:encoding',
```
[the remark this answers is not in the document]
tmathern: Checked - they are waiting for us to finish the change to reflect it on AEM side.

lib/rendition.js
```diff
@@ -42,17 +102,158 @@ class Rendition {
...
+        } 
+        // fallback if things are missing ------------------------------------------------------
+        if(!this._encoding || !this._mime){ // fallback to using file-type lib (won't set encoding)
+            try { 
+                await commandExists('file'); // throws if `file` is not found
+                const fileCmdResult = await getMimeInformationFromCommand(this);
+                this._mime = this._mime || fileCmdResult.mime;
+                this._encoding = this._encoding || fileCmdResult.encoding;
+            } catch(err){
+                console.log('`file` command not found', err); // err should be empty
```
alexkli, line 150: This can also end up if `file` fails for some reason, so the log message is not always true. You can just change it to `file command failed:` and the logging of the `err` will tell what the problem was, including exit code 127 for not found IIRC.

lib/rendition.js
```diff
@@ -42,17 +102,158 @@ class Rendition {
...
+                console.log('Trying to find mimetype based on rendition file analysis (magic numbers)');
```
alexkli, line 151: The same was tried already, so maybe clarify:
```
Falling back to detecting mimetype using `file-type` library
```

lib/rendition.js
```diff
@@ -42,17 +102,158 @@ class Rendition {
...
+
+    /**
+     * Returns just the encoding
+     */
+    async encoding() {
+        if(!this._encoding){
+            await this.setMimeInformation();
+        }
+
+        if(this._encoding === "charset=binary"){
```
alexkli, line 178: If a client/worker sets it explicitly, we should take it for granted, including this case. So I think this filter (from the `file` result) should happen in the `getMimeInformationFromCommand()`/`detectTypeUsingFileCommand()` function where it "originates".

Not sure if `file-type` could also report a binary charset? If yes, handle it for both in the general detect function.

package.json
```diff
@@ -16,33 +16,35 @@
     "@adobe/httptransfer": "^1.0.1",
     "@adobe/metrics-sampler": "^1.0.0",
     "@adobe/node-fetch-retry": "^1.0.1",
-    "ajv": "6.12.2",
+    "ajv": "6.12.3",
     "clone": "^2.1.2",
+    "command-exists": "^1.2.9",
```
alexkli, line 21: As mentioned above, I think we don't need this dependency.

package.json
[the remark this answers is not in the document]
alexkli: Yes, but would be nice to add the `-r=lcov` for support in vs code coverage gutters.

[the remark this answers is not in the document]
alexkli: Also I believe the `-- -v` is just for development and is meant to be removed before the PR is merged?

test/rendition.content.test.js
```diff
@@ -0,0 +1,150 @@
...
+/* eslint mocha/no-mocha-arrows: "off" */
+
+'use strict';
+
+const assert = require('assert');
+const Rendition = require('../lib/rendition.js');
+
+// WARNING: filesystem is not mocked here, so content-type identification can be tested with known files.
+
+describe("rendition.js - content types", () => {
```
alexkli, line 23: Do these tests depend on the existence of the `file` command?

I think we should mock it, and be able to test both file and file-type fallback cases.

lib/rendition.js
```diff
@@ -42,17 +102,158 @@ class Rendition {
...
+     *      - mime-type + encoding for other (encoding for txt files)
+     */
+    async contentType() {
+        let contentType;
+
+        const encoding = await this.encoding();
+        const mimetype = await this.mimeType();
+
+        if(encoding){
+            contentType = `${mimetype}; ${encoding}`;
```
alexkli, line 196: This must be `charset=${encoding}`

The encoding is just the encoding, not the string `charset=<encoding>`.

This is needed for:
1. Metadata.ENCODING must not include `charset`
2. The client when calling `setContentType()` would do:

```
setContentType("text/html", "utf-8")
```

not this
```
setContentType("text/html", "charset=utf-8")
```

This means parsing it from the `file` command result, but that's ok.

test/rendition.test.js
```diff
@@ -156,4 +172,171 @@ describe("rendition.js", () => {
...
+        rendition.setMimeInformation("txt/plain", "charset=ascii");
+
+        const mime = await rendition.mimeType();
+        assert.strictEqual(mime, "txt/plain");
+
+        const encoding = await rendition.encoding();
+        assert.strictEqual(encoding, "charset=ascii");
+
+        const contentType = await rendition.contentType();
+        assert.strictEqual(contentType, "txt/plain; charset=ascii");
```
alexkli, line 224: As mentioned above, the client would set the encoding value `ascii` here, not `charset=ascii`.

lib/shell/shellscript.js
```diff
@@ -44,10 +44,11 @@ class ShellScriptWorker {
     async processWithScript(source, rendition) {
         console.log(`executing shell script ${this.script} for rendition ${rendition.id()}`);
 
-        const errorFile = prepareErrorFile(rendition.directory);
+        const errorFile = await prepareErrorFile(rendition.directory);
+        rendition.mimeInfoPath = await prepareMimeInfoFile(rendition.directory, rendition.id());
```
alexkli, line 48: Abstraction/encapsulation: The `mimeInfoPath` should be read by the shell script worker itself, and it would call `rendition.setContentType()` accordingly. The Rendition class should not have to know about the logic of using a file here and its format.

lib/rendition.js
```diff
@@ -42,17 +102,158 @@ class Rendition {
...
+            this._mime = mime;
+        }
+
+        // explicitly set encoding would get priority
+        if(encoding){ // to handle case when this.encoding would have a default value
+            this._encoding = encoding;
+        }
+
+        // shellscript mostly ------------------------------------------------------
+        if(this.mimeInfoPath && await fs.pathExists(this.mimeInfoPath)){ // use the path containing the file with mime information
```
alexkli, line 132: This logic should be in the shellscript.js (see comment over there).

lib/rendition.js
```diff
@@ -42,17 +102,158 @@ class Rendition {
...
+            let mimeInfoContent = await fs.readFile(this.mimeInfoPath);
+            mimeInfoContent = mimeInfoContent.toString();
+            const mimeInfo = mimeInfoContent.split(";");
```
alexkli, line 135: We should probably use a library for that, we use [content-type](https://www.npmjs.com/package/content-type) in core.

lib/shell/shellscript.js
```diff
@@ -110,13 +117,14 @@ function newScriptError(err, errorFile) {
...
 
-function setVariables(rendition, source, errorFile){
+function setVariables(rendition, source, errorFile, mimeFile){
     // inherit environment variables
     const env = Object.create(process.env || {});
 
     env.source = env.file = stripAnsi(source.path); // escape because user provided
     env.rendition = stripAnsi(rendition.path);
     env.errorfile = errorFile;
+    env.mimefile = mimeFile;
```
alexkli, line 127: `mime` itself isn't a commonly used term. Also this file is supposed to include the full content type.

Maybe name this `contenttypefile` or just `typefile` (shorter).

lib/shell/shellscript.js
```diff
@@ -59,12 +60,18 @@ class ShellScriptWorker {
...
-function prepareErrorFile(directory) {
+async function prepareErrorFile(directory) {
     const errDir = path.resolve(directory, "errors");
-    fs.mkdirSync(errDir, {recursive: true});
+    await fs.mkdirs(errDir, {recursive: true});
     return path.resolve(errDir, "error.json");
 }
 
+async function prepareMimeInfoFile(directory, renditionId) {
+    const mimeDir = path.resolve(directory, "mimeinfo");
```
alexkli, line 70: Maybe instead of extra directories have one common directory for these special files: `metadata`

And put the error and type file in there.

Since clients are supposed to use the path from the env var we should be able to make this change for the error file.

lib/shell/shellscript.js
```diff
@@ -59,12 +60,18 @@ class ShellScriptWorker {
...
+    await fs.mkdirs(mimeDir, {recursive: true});
+    return path.resolve(mimeDir, `mimeinfo-${renditionId}.txt`);
```
alexkli, line 72: At this point the shell script worker only supports one rendition at a time, so we don't need to support multiple renditions at a time.

And same "mime" terminology remark.

Suggest to name `type.txt` or `contentype.txt`. And remove this file/the `metadata` folder after each rendition.

lib/rendition.js
```diff
@@ -42,17 +102,158 @@ class Rendition {
...
+        return this._sha1;
     }
 
-    sha1() {
-        return this.metadata[RenditionMetadata.REPO_SHA1];
+    exists() { // see and read the file
+        return fs.existsSync(this.path);
+    }
+
+    async setMimeInformation(mime=null, encoding=null){
```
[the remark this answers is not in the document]
tmathern: To detail:
- `setContentType`blindly trusts the client, so I wouldn't check for the presence of `charset=`in the encoding param. Or do we want the friendliness of removing it if it is here? (I think we shouldn't - but double checking). 
- `detectContentType` was going to happen anyway in refactoring. I wanted feedback over general concepts/API before refactoring, and you confirmed it's OK to set mime and encoding together (makes life so much easier).

lib/rendition.js
```diff
@@ -12,28 +12,88 @@
...
+const crypto = require('crypto');
+const commandExists = require('command-exists');
+
+const imageSize = util.promisify(require('image-size'));
+const FileType = require('file-type');
+
+const exec = util.promisify(require('child_process').exec);
 
 const RENDITION_BASENAME = 'rendition';
+const DEFAULT_CONTENT_TYPE = 'application/octet-stream';
```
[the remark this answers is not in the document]
tmathern: done.

lib/rendition.js
```diff
@@ -42,17 +102,158 @@ class Rendition {
...
+
+            if(mimeInfo && mimeInfo.length === 2){ // we got something in an expected format
+                this._mime = mime || mimeInfo[0].trim() || DEFAULT_CONTENT_TYPE;
+                this._encoding = encoding || mimeInfo[1].trim();
+            }
+        } 
+        // fallback if things are missing ------------------------------------------------------
+        if(!this._encoding || !this._mime){ // fallback to using file-type lib (won't set encoding)
+            try { 
+                await commandExists('file'); // throws if `file` is not found
```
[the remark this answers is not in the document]
tmathern: Some context: My idea by checking was to avoid running a "useless" (failing) exec. So I double check that lib's code and it falls back anyway to trying it with exec.

Also, we're sure we only need this working on *nix, no WIndows support needed for this for now. 

Because of those 2 reasons, agreed, I'll remove the commandExists check.

lib/rendition.js
```diff
@@ -42,17 +102,158 @@ class Rendition {
...
+                const fileCmdResult = await getMimeInformationFromCommand(this);
```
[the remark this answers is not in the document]
tmathern: Irrelevant after removal of `commandExists`.

lib/shell/shellscript.js
tmathern: Can we do this in a second step? 
So here I don't need to add the change to the error handling and can keep the PR focused.
Also, since changing error handling may be a bit touchy, I would prefer to do this folder change after the mime detection.
If it is OK, I can create a JIRA story for next sprint. Wdyt?

lib/rendition.js
tmathern: ACK.

lib/shell/shellscript.js
tmathern: I'll rename to `typefile`.

lib/rendition.js
tmathern: Double-check:

You mean after the processing (line https://github.com/adobe/asset-compute-sdk/blob/e866028b9d3638fa9caad455037cef8bbae0b379/lib/shell/shellscript.js#L55), right?

So the shellscript still can write its file and there is no usage change on that end.

lib/shell/shellscript.js
tmathern: `type.txt` it is.

lib/rendition.js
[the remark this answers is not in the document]
alexkli: 1. setContentType expects no charset= so does not have to strip it
2. ✌️

lib/rendition.js
alexkli: yes, after the shell script was invoked in `processWithScript()` it checks if the type file exists and parses it, and calls rendition.setContentType(). just like the error file is also handled in there (inside newScriptError()).

lib/shell/shellscript.js
alexkli: Then you could put the new type file in the errors directory now, so it’s the same directory already.

package.json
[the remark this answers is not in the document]
tmathern: `-r=lcov` makes the command fail for me. Could you tell me if you installed a package globally? Or only vs code coverage gutters?

And `-- -v` is wanted here, to display log outputs.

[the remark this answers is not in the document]
alexkli: Not sure, compare with https://github.com/adobe/aio-cli-plugin-asset-compute/blob/master/package.json#L83 which runs fine in CI.

test/rendition.content.test.js
```diff
@@ -0,0 +1,189 @@
...
+
+const assert = require('assert');
+const Rendition = require('../lib/rendition.js');
+
+const rewire = require("rewire");
+const rewiredRendition = rewire('../lib/rendition.js');
+
+// WARNING: filesystem is not mocked here, so content-type identification can be tested with known files.
+
+describe("rendition.js - content types", () => {
```
alexkli, line 26: Some content type related tests are over in `rendition.test.js` such as `handles gracefully reading incomplete data for mime+encoding`. I would have them all in one place, e.g. inside `rendition.tests.js`.

tmathern: Oh, that was on purpose. I had some issues because of the mocked file system, so the `.content`-tests do not use mockFs. I noticed that mockFs `.restore` is actually not that immediate and I had flaky tests when tests without mockFs were in the same file. It's more a workaround than a true fix for mockFs.

lib/shell/shellscript.js
```diff
@@ -44,25 +48,53 @@ class ShellScriptWorker {
...
+    // TODO: JIRA story to move errors content into a new `metadata` folder instead of `errors`
     const errDir = path.resolve(directory, "errors");
-    fs.mkdirSync(errDir, {recursive: true});
-    return path.resolve(errDir, "error.json");
+    await fs.mkdirs(errDir, {recursive: true});
+
+    const errFile = path.resolve(errDir, "error.json");
+
+    // Folder structure has activationid as root, so concurrent executions should not collide
+    // so no need for: const typeFile = path.resolve(mimeDir, `type-${rendition.id()}.txt`);
```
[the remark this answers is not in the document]
tmathern: I added it because I actually double-checked while writing the code.

test/rendition.content.test.js
```diff
@@ -0,0 +1,189 @@
...
+        assert.strictEqual(metadata["tiff:imageHeight"], 288);
+        assert.strictEqual(metadata["dc:format"], "image/jpeg");
+        assert.ok(metadata["repo:encoding"] === undefined);
+    });
+
+    it('tests fallback from file command to file-type lib', async function () {
+        const failingFileCommand = async () => {
+            throw new Error('file command failure simulation');
+        };
+        rewiredRendition.__set__("getContentTypeInformation", failingFileCommand);
```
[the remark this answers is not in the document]
tmathern: I have nothing about doing white box testing when writing unit tests, even testing internal functions if some executions paths are hard to achieve. 
The thing I don't like about temporary overwriting functions/processes/libraries is that it is easy to forget they are overwritten in the test, and need to be restored. Especially when overwriting things broadly reusable, like `child_process`.
That is actually why I prefer rewire: when rewired, the rewired item under test is expected to not necessarily be in a good state. It's white box testing, and subject to change, sure. But if I would change the behavior, I probably also would want the test to fail so that I attend to the effects of the change.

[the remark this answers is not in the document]
alexkli: > temporary overwriting functions/processes/libraries is that it is easy to forget they are overwritten in the test
They should be setup and restored for each test, e.g. using `afterEach()`.

test/rendition.content.test.js
alexkli: > I noticed that mockFs .restore is actually not that immediate

That can't be. A missing `await` maybe?

Another solution is to make mockFs more selective on each test, not "blindly" on all.

tmathern: `make mockFs more selective on each test that requires it` <= we had issues before (not this PR) and tried this. This made it worse. It's something we will need to investigate.

alexkli: But that must have been other specific reasons of other projects and their tests.

Generally, we have been doing a lot of prepration in `beforeEach()` which applies to all tests. But this becomes problematic the more tests you add that might want a different environment - this here is one example. This apples to all kinds of setup, not just mockFs.

What I learned is better is to put all the setup in the individual tests. If the setup code is reused, put it in a shared function that those tests that need it can reuse it.

alexkli:

PR #17, opened 2020-07-07
Removal of compat layer (1/2)

tmathern:
## Description

Removes the compatibility layer we used to transition workers after sdk refactoring (it has been deprecated for some months, now). Time to remove it!

Fixes #NUI-483.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] I have signed the [Adobe Open Source CLA](https://opensource.adobe.com/cla.html).
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #18, opened 2020-07-07
BREAKING-RELEASE: Removes error.js and puts error exports in index.js (2/2) 

tmathern:
## Description

Export errors in index.js (instead of errors.js).
Will need to update workers with this sdk change once the export is moved.
This should be considered a breaking-change... Wdyt @alexkli ?

Fixes # [NUI-651](https://jira.corp.adobe.com/browse/NUI-651)

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:


tmathern:
@alexkli Actually, I've got another proposale: WHen I did the PR, I didn't know if/when https://github.com/adobe/asset-compute-sdk/pull/17 would get a review+approval. Since this one is a breaking change anyway, I would merge https://github.com/adobe/asset-compute-sdk/pull/17 and this one and _then_ do the breaking-change release.

We don't really need to keep the errors export, I already have PRs ready to update impacted components. It's a matter of bumping the version number of the sdk dependency and re-running the tests. 

Ok if we handle it that way? So we totally get rid of transition code.

tmathern:
Failing tests currently are expected, as the workers will need the new SDK.

alexkli:
@tmathern Another breaking change is NOT a good reason to add more breaking changes. Clients would still have to make the change, whereas otherwise they wouldn't. Look at clients, not just the semver version bump.

Leaving in the `errors.js` to keep backwards compatible is not really a problem since it's not duplicated code, but just passes through the error classes from commons.

The point of NUI-651 was to export the errors in the root since that makes for a better API usage, NOT to get rid of errors.js at all costs.

tmathern:
I get you're not happy, but I did not blindly do the changes and asked multiple time about feedback. I also agree at limit scope change in breaking change. But one breaking change per major? Not necessary. Other libraries and components have more than one breaking change per major bump too. 

I looked at client (users) of the SDK, and most of them (internal workers) already have the change and are released. I actually prepared those changes in PRs and branches BEFORE merging the sdk. And retriggering tests after release. Why re-test? In case I would have had to rollback/unpublish. In the end, I assure it was not blindly done, but I understand your frustration and remarks (and mostly agree). 

I get your point of leaving it in, but here the impact and risk is low (we have a lot of test coverage, which also properly flagged the failing elements). And we knoe what to change when, And multiple tests ran on too!

Also, if the point is about the generators and templates, the generator still generates the proper worker for the template. There is no version mismatch and we can keep them in lockstep. I'll actually go update them (but, as I said, currently, they are in lockstep and not in a dangling state). 

Fine, next time I will do 2 major bumps. But here, given the proximity, the tests, and the low risks, I respectfully disagree and say it was OK to do it that way.

alexkli:

PR #20, opened 2020-07-14
Make error less confusing

tmathern:
## Description

`file` command is not available everywhere, so it may fail.
This is to change the corresponding error message to something less confusing, since the exception is properly handled in the code.
(Found during hackathon. Since previously stacktrace was logged, people believed this was the reason workers were crashing... but it wasn't since, when you continue reading logs, you see the error is handled. Hence the change to clarify that).

Fixes # (issue)

## Types of changes

## Checklist:


lib/rendition.js
```diff
@@ -82,9 +82,9 @@ async function detectContentType(renditionPath) {
         const fileCmdResult = await getContentTypeInformation(renditionPath);
         localMimeType = fileCmdResult.mime;
         localEncoding = fileCmdResult.encoding;
-    } catch (err) { // fallback to using `file-type lib` (won't set encoding)
-        console.log('`file` command failed', err); // err should be empty
-        console.log('Trying to find mimetype based on rendition file analysis (magic numbers)');
+    } catch (err) { /* eslint-disable-line no-unused-vars */
+        // fallback to using `file-type lib` (won't set encoding)
+        console.log('`file` command failed. Trying to find mimetype based on rendition file analysis (magic numbers)');
```
alexkli, line 87: Same comment as before: this log statement is confusing, since this is exactly what `file` is doing as well:

```
Trying to find mimetype based on rendition file analysis (magic numbers)
```

Either just drop it or say "falling back to `file-type` library".

tmathern: ack

alexkli:

PR #21, opened 2020-07-15
FEATURE-RELEASE: Send IO events for all the unfinished renditions on timeout

jdelbick:
## Description

- send  a `rendition_failed` IO event for each incomplete rendition (`rendition.eventSent === false`) with error reason `GenericError` on action timeout
- abort processing immediately by calling `process.exit(124)` (let me know if a different exit code makes more sense)
- copied [timeout logic](https://github.com/adobe/node-openwhisk-newrelic/blob/master/lib/metrics.js#L38-L44) from `node-openwhisk-newrelic`. This library already [sends metrics on timeouts](https://github.com/adobe/node-openwhisk-newrelic/blob/master/lib/newrelic.js#L158-L186) and offers a callback function to run before timing out. I tried to piggy back off of the `actionTimeoutMetricsCb()` function here so I wouldn't need to duplicate the logic, but this cb does not have access to the current state of `AssetComputeWorker` which it needs to send IO events. Also the [`sendMetricsOnActionTimeout`](https://github.com/adobe/node-openwhisk-newrelic/blob/master/lib/newrelic.js#L92) timeout is created in the constructor of the new relic events client and defined in a few different spots in the code, [wrapping the action](https://github.com/adobe/asset-compute-commons/blob/master/lib/wrap.js#L50) or [AssetComputeWorker constructor](https://github.com/adobe/asset-compute-sdk/blob/master/lib/worker.js#L65) and can still be overwritten by [passed in metrics](https://github.com/adobe/asset-compute-commons/blob/master/lib/wrap.js#L52) so it felt too fragile to depend on defining `actionTimeoutCb`.

to avoid the duplicated code logic + risk of timeout buffers in one library adjusting in the future, I could disable the action timeout in `openwhisk-newrelic` and take care of it in the sdk. so the setTimeout defined in the sdk will both send io events and send a timeout metric on action timeout. Let me know what you think.

Fixes # https://jira.corp.adobe.com/browse/NUI-663

I tested this with a worker I made that just sleeps forever and deployed it with an openwhisk action timeout of 15s. Here are the results I got in the devtool:
Screen Shot 2020-07-14 at 4 30 59 PM

## Types of changes

## Checklist:


lib/worker.js
```diff
@@ -330,12 +344,39 @@ class AssetComputeWorker {
...
+                                errorReason: Reason.GenericError,
+                                errorMessage: "Timeout",
+                                activationIds: this.params.customWorker ? [process.env.__OW_ACTIVATION_ID] : undefined
+                            });
+                            rendition.eventSent = true;
+                        }
+                    }
+                }
+                // Abort processsing. Process ended abnormally by timeout
+                process.exit(TIMEOUT_EXIT_CODE);
```
alexkli, line 366: Killing the process here means the container can't be reused (as it can't accept any requests/activations anymore, even if the current one finishes in time). However the goal is actually the opposite: "avoid having the container killed" (see [NUI-663](https://jira.corp.adobe.com/browse/NUI-663)).

What we should do is see if the rendition callback can be aborted... maybe that needs a new api/spi for workers to implement so they can kill their child processes (etc.).

And then the action should return a result when this timeout handler is called. Returning a result is key for openwhisk to not see this as timeout. Technically this could be done through a promise etc. in the main worker code, which finishes when either the overall processing loop is done (all renditions) or this timeout is called and sets something. A bit tricky but should be possible.

Of course if we cannot abort the current callback, but we return, then this container will get new activations while there might be still some async operations (in nodejs) or a child process churning away. Not sure about the risk here.

test/testutil.js
```diff
@@ -82,6 +82,7 @@ function nockIOEvent(expectedPayload, status=200) {
     return nock("https://eg-ingress.adobe.io")
         .post("/api/events", body => {
             const payload = parseIoEventPayload(body.event);
+            console.log('event payload', payload);
```
alexkli, line 85: Log message should be removed.

test/api.test.js
```diff
@@ -535,6 +545,74 @@ describe("api.js", () => {
...
+            });
+
+            const main = worker(async function(source, rendition) {
+                fs.writeFileSync(rendition.path, testUtil.RENDITION_CONTENT);
+                await sleep(500);
+                console.log('waiting...');
+                return Promise.resolve();
+            });
+            assert.equal(typeof main, "function");
+            process.env.__OW_DEADLINE = Date.now() + 300;
```
alexkli, line 569: You can probably set this down to 100 ms or even 50 ms to make the test faster with the same result.

test/api.test.js
```diff
@@ -535,6 +545,74 @@ describe("api.js", () => {
...
+            const main = worker(async function(source, rendition) {
+                if (rendition.index === 2) {
+                    console.log('waiting...');
+                    await sleep(5000);
+                }
+                fs.writeFileSync(rendition.path, testUtil.RENDITION_CONTENT);
+                return Promise.resolve();
+            });
+            assert.equal(typeof main, "function");
+            process.env.__OW_DEADLINE = Date.now() + 7000;
```
alexkli, line 602: Same here, 7 seconds is very long, why not 100ms (and adapting the other sleep value accordingly)?

lib/worker.js
```diff
@@ -330,12 +344,39 @@ class AssetComputeWorker {
...
+            async () => {
+                console.log(`Action is about to timeout in ${timeUntilTimeout()}ms Sending IO events before timeout`);
+                // ensure failure events are sent for any non successful rendition before timeout
+                if(this.renditions){
+                    for (const rendition of this.renditions) {
+                        if (!rendition.eventSent) {
+                            await this.events.sendEvent(EVENT_RENDITION_FAILED, {
+                                rendition: rendition.instructionsForEvent(),
+                                errorReason: Reason.GenericError,
+                                errorMessage: "Timeout",
```
alexkli, line 358: Maybe add a bit more text: "Timed out without result after XXX seconds/minutes"

Leads to an interesting question: time from start of worker? That particular rendition might not have been started yet (another one or the download could have been the blocking issue).

lib/worker.js
```diff
@@ -330,12 +344,39 @@ class AssetComputeWorker {
...
+                                activationIds: this.params.customWorker ? [process.env.__OW_ACTIVATION_ID] : undefined
+                            });
```
alexkli, line 360: We also need to send the corresponding rendition failed metrics (`handleError()`). See https://github.com/adobe/asset-compute-sdk/blob/master/lib/worker.js#L317-L329

(This somewhat makes the `timeout` metric duplicated. But it doesn't hurt to have the two. We can always disable later once we are in production and actually see the data in NR and think it's superfluous).

lib/worker.js
jdelbick: Ah I see. I misinterpreted the ticket.

> And then the action should return a result when this timeout handler is called. Returning a result is key for openwhisk to not see this as timeout. Technically this could be done through a promise etc. in the main worker code, which finishes when either the overall processing loop is done (all renditions) or this timeout is called and sets something. A bit tricky but should be possible.

^I think we can add that promise inside [here](https://github.com/adobe/asset-compute-sdk/blob/master/lib/worker.js#L108-L123). I will work on that now. 

I will also explore the idea of having an api/spi to abort rendition callbacks.

tmathern: p-cancelable looks good, indeed (just make sure both the cancellation code and the event sending can be sent before the timeout).

test/api.test.js
```diff
@@ -535,6 +545,74 @@ describe("api.js", () => {
...
+
+            testUtil.assertNockDone();
+            await MetricsTestHelper.metricsDone();
+            MetricsTestHelper.assertArrayContains(receivedMetrics, [{
+                eventType: 'timeout'
+            },{
+                eventType: "activation"
+            }]);
+            assert.equal(processSpy.calledOnce, true, "did not call process.exit(124) on timeout");
+        }).timeout(10000);
```
tmathern, line 614: Also, if you reduce the wait time, maybe you can (at least) reduce the timeout value here, or even remove it.

test/api.test.js
jdelbick: The reason I chose 7s here is that I am testing a timeout in the middle of processing. The goal of this test is to process 2 renditions successfully and the third one times out. The test should send 2 rendition created events and 1 rendition failed event. (The reason i used 3 renditions instead of 2 is that the [multiple rendition params function](https://github.com/adobe/asset-compute-sdk/blob/a0df461eedd7fb55d760d04b19f3ab0e12cd20ae/test/testutil.js#L187) from the test utils had 3 renditions). To successfully do this, I need the timeout to be > 5s since that is the [timeout buffer](https://github.com/adobe/asset-compute-sdk/blob/a0df461eedd7fb55d760d04b19f3ab0e12cd20ae/lib/worker.js#L35). If the action deadline is in <5s, the action will start its timeout process during the first rendition.

One thing I can try is using sinon fake timers to skip ahead in time instead of sleeping for 5s so the test will not actually take 5+ seconds to run

lib/worker.js
alexkli: I mean maybe `process.exit()` is the best and safest solution we can do right now. Or maybe we offer both options, since "canceling" the processing properly is dependent on the worker implementation - we probably cannot guarantee it here.

So one model might be:
1. by default do process.exit() => safe
2. if the worker opts in, support canceling & container reuse

For 2: ideally using an api that automatically opts in, e.g. register an "abort" callback that is supposed to kill any child process etc., and once that is registered we don't do the process.exit.

Would be good to understand how good the cancelable promise approach can work and how much we'd have to refactor for it.

If too difficult, keep this PR to 1., and add 2. later (which is possible if it's opt in).

wdyt?

test/api.test.js
alexkli: > I need the timeout to be > 5s since that is the timeout buffer. If the action deadline is in <5s, the action will start its timeout process during the first rendition.

I see. I am ok with that then (if fake timers don't work out).

lib/worker.js
```diff
@@ -63,6 +71,12 @@ class AssetComputeWorker {
 
         this.events  = new AssetComputeEvents(this.params);
         this.metrics = params.metrics || new AssetComputeMetrics(params);
+
+        // set timeout to send events before action timeout
+        if (!process.env.DISABLE_IO_EVENTS_ON_TIMEOUT) {
```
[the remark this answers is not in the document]
alexkli: First look at the config option. The default for a boolean option should always be false/ not set. This is the case here since IO events on timeout are to be sent by default. So the option must be "DISABLE=true".

This does lead to a double negative here in the code, but this is worth it for the consumer.

lib/worker.js
jdelbick: I just refactored the main run loop to use promises so we can gracefully return if the action is about to timeout. The next step is trying/testing out the p-cancelable library to see if we can abort the renditionCB. I will spend some time exploring this. If I find it is unreliable, I will switch back to `process.exit()` for now

lib/worker.js
jdelbick: Currently it is the time from start of worker. Maybe to make it clearer we can say: "Worker timed out without result after XXX seconds/minutes"

lib/worker.js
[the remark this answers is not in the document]
jdelbick: Yep, this is the same reason we defined it this way for [`node-openwhisk-newrelic`](https://github.com/adobe/node-openwhisk-newrelic/blob/master/lib/newrelic.js#L91).

This env var is really only meant for unit testing anyways

lib/worker.js
jdelbick: Ok so I played around with the current behavior and this is what I found. A quick summary of the current behavior:
if the worker is about to timeout (<5s until action timeout), it will send `rendition_failed` events then end the action successfully. It will set `rendition.eventSent = true` right away so there is no race condition in case the rendition finishes before it times out. The same container can be used again.

I tried two different scenarios where the worker timeout is set to 15s:
1. worker sleeps for 30s:
it is not expected here that the worker will complete processing in time before timeout. In testing, it does not complete processing in time. It sends a rendition_failed event for a timeout, does not end up uploading a rendition, and the action finishes successfully. (Same docker image is used in subsequent process calls.)
2. worker sleeps for 10s:
it is expected that the worker will finish before the action actually times out but will be canceled before it finishes processing. The sdk will start "shutting the action down" after 10s, which should be just before the worker finishes processing when you take into account download + set up time.
In testing, it sent a rendition_failed event for timeout but still uploaded the asset in time before the action finished. 


I spent some time trying to figure out how we can "abort processing". The short conclusion is, we can abort processing in all parts of the code that we control (we can add a variable that checks at each step of processing uploading/downloading,etc to see if we went into "timeout mode"). However, we will never be able to control what is inside the custom workers `renditionCallback` and there is no way in javascript (besides process.kill, process.abort, etc.) to kill this cb once it has started since javascript is single-threaded by nature.

So this leaves a few ideas,
1. we keep it as I currently implemented it (describe at the beginning of this comment) and add a boolean variable `this.actionTimedout=true` and check it before each step of the sdk processing.

   My only concern with this is after the action ends due to timeout and returns successfully, what happens in the case 
 that the `renditionCallback` is still running? (first scenario I tried out).  I assume the node process gets killed after the  action returns successfully so there is no `renditionCallback` running in the background of the container...but I cannot confirm this. I do know the action never ends up uploading a rendition and the container gets used again in subsequent calls, which is good.

2. we kill the process when the action is about to timeout (basically I revert my changes made in the last commit). This way we know for sure there are no background processes still running in the container

lib/worker.js
alexkli: 👍

lib/worker.js
alexkli: I think you convinced me that 2. `process.exit()` is the safe route. The code also becomes more complex with option 1. looking at the latest commits, which might hurt our maintainability of it going forward (higher risk of new bugs).

Maybe if we see an issue with those self-killed containers in the future we can come back and look at a solution towards option 1.

> <5s until timeout

Maybe make the buffer longer. We might have to send a couple of IO Events and they could fail and retry potentially (say the reason for the worker timing out was slow network etc.). Not sure, maybe 30 seconds? Our shortest worker timeouts are 5 minutes I believe...

lib/worker.js
jdelbick: Done, see below. Since we will not really know what part of the processing the code is in when it times out, it is hard to determine which metrics to add (processingDuration etc.) so all I added was `duration` for now

lib/worker.js
jdelbick: Alright, I just reverted it back to the older commit (and fixed the previous comments on the review).

As far as timeout buffer, I looked at some new relic data to get some information on how long actions run:
https://insights.newrelic.com/apps/accounts/2045220/asset-compute-production/dashboards/969243?query=SELECT%20median(duration),%20average(duration),%20max(duration)%20FROM%20activation%20WHERE%20namespace%3D%27126848_50549%27%20WHERE%20actionName%20NOT%20IN%20(%27api-process%27,%20%27api-register%27,%20%27core%27,%20%27auth%27)%20since%201%20month%20ago
(median seems to give the 50th percentile as an object and not a number)
```
Action duration over the past month in seconds:
Median: 1.2994146
Avg Duration: 3.55
Max Duration: 3,659
```

I would be hesitant to say 30s buffer time since the default timeout for a runtime action is 60s, and when writing a custom worker (as I observed during the hackathon) most do not change the default value if they don't see a need for it. It would be kind of annoying/confusing when developing to see the action begin to timeout halfway through the time specified. Maybe we do 15 seconds as the timeout buffer?

tmathern: I always love using the Pareto distribution (https://en.wikipedia.org/wiki/Pareto_distribution and/or https://en.wikipedia.org/wiki/Pareto_principle). Meaning: for many events, roughly 80% of the effects come from 20% of the causes.

I would say give 80% of the time to the worker, keep 20% for the timeout handling (means the worker gets 48s out of 60s). Wdyt?

jdelbick: Yeah I think that is a good idea.

I would like to calculate the timeout buffer based on max action duration (timeout field in the manifest yaml) but I don't know how we would have access to that besides looking at `process.env.__OW_ACTION_DEADLINE`. I worry the environment variable that is just the time until the action time sout could be off if for some reason the worker does some processing before calling the sdk, etc.

jdelbick: However... a full minute feels like a long time to be "timeout handling"...especially since the worker timeout has a grace period of about ~20s.

I went with 15s. Let me know your thoughts. We can always adjust this in the future.

tmathern: ack.

pheenomenon:
@jdelbick few questions:
>(Same docker image is used in subsequent process calls.)

How do you verify this?

Also, I was wondering if this change also effects the multi-part rendition processing ? thoughts?

lib/worker.js
alexkli: @jdelbick regarding median in NRQL: use `percentile(duration, 50)` instead of `median()`, then it works.

alexkli: 1. Here is a [query with some 99.x percentiles](https://insights.newrelic.com/apps/accounts/2045220/asset-compute-production/dashboards/969243?query=SELECT%20percentile(duration,%2050)%20AS%20%27median%27,%20percentile(duration,%2099),%20percentile(duration,%2099.9),%20percentile(duration,%2099.94),%20max(duration)%20FROM%20activation%20WHERE%20namespace%3D%27126848_50549%27%20WHERE%20actionName%20LIKE%20%27worker-%25%27%20AND%20actionName%20!%3D%20%27worker-pie-large%27%20AND%20actionName%20!%3D%20%27worker-tika%27%20since%2090%20days%20ago) (and excluding worker-pie-large & worker-tika which are the two I know that have > 5 min timeout).

   Playing around with the percentiles:
   - `99.998%tile` is 264 seconds
   - `99.999%tile` is 296 seconds

   Meaning `< 0.002%` of requests would now fail earlier instead of finishing. (At least I think the query shows _successful_ ones that did not timeout...). Should be ok I guess.

2. How many retries do we get in 15 seconds? If the socket/network timeout is 30 sec, then we would not get any retry in this case. That's why I was thinking about more.

   Default timeout in workers: we could increase this in the generator to say 5 min. At least have a value in the manifest so that it's easy to find and change. But I can also see that Runtime might want us to keep it at 1 min max to avoid too many long running actions where it's not necessary.

   Other than the retry, we only need a fairly fixed amount of time to finish the action (send 1 to few io events), so a general percentage would lead to unneeded long durations for longer timeouts.

   If we need it, we could have a staggered setup: 10 sec for 1 minute, 30 sec for 5 min, 1 min (max?) for 10min+...

   But I am ok with 15 sec for now.

lib/worker.js
```diff
@@ -25,11 +25,14 @@ const { Sampler } = require('@adobe/metrics-sampler');
 const { metrics: cgroupMetrics, cpu} = require('@adobe/cgroup-metrics');
 
 const CLEANUP_FAILED_EXIT_CODE = 231;
+const TIMEOUT_EXIT_CODE = 124;
```
alexkli, line 28: Any reason for this number in particular?

It seems we have very random numbers here, 231, 124... maybe keep them more systematic?

lib/worker.js
```diff
@@ -330,12 +345,47 @@ class AssetComputeWorker {
         }
     }
 
+    sendEventsBeforeActionTimeout() {
+        return setTimeout(
+            async () => {
+                console.log(`Action is about to timeout in ${timeUntilTimeout()}ms. Sending Adobe I/O events before timeout.`);
```
alexkli, line 351: Maybe say this instead:

```
... Sending rendition_failed events before timeout.
```

lib/worker.js
```diff
@@ -330,12 +345,47 @@ class AssetComputeWorker {
...
+                if(this.renditions){
+                    for (const rendition of this.renditions) {
+                        if (!rendition.eventSent) {
+                            await this.events.sendEvent(EVENT_RENDITION_FAILED, {
+                                rendition: rendition.instructionsForEvent(),
+                                errorReason: Reason.GenericError,
+                                errorMessage: errorMessage,
+                                activationIds: this.params.customWorker ? [process.env.__OW_ACTIVATION_ID] : undefined
+                            });
+                            rendition.eventSent = true;
```
alexkli, line 363: I think in this case we should have the `rendition.eventSent = true;` _before_ the `await this.events.sendEvent()` to ensure another concurrently finishing processing does not send the event.

lib/worker.js
```diff
@@ -330,12 +345,47 @@ class AssetComputeWorker {
...
+                        }
+                    }
+                }
+
+                await this.metrics.handleError(new GenericError(errorMessage), {
```
alexkli, line 368: Are we still sending the `timeout` metric? I think we must ensure this somehow as we would fully consider this path as a timeout. Maybe send it manually here instead of the error metric.

Currently it happens here: https://github.com/adobe/node-openwhisk-newrelic/blob/af62c5ce3e3c465655e624b73d81d290ed39e03e/lib/newrelic.js#L180

test/api.test.js
```diff
@@ -535,6 +546,68 @@ describe("api.js", () => {
...
+            assert.equal(typeof main, "function");
+            process.env.__OW_DEADLINE = Date.now() + 100;
+
+            await main(testUtil.simpleParams({noEventsNock:true, noPut:true}));
+            testUtil.assertNockDone();
+            await MetricsTestHelper.metricsDone();
+            MetricsTestHelper.assertArrayContains(receivedMetrics, [{
+                eventType: 'error',
+                location: 'test_action_timeout'
+            }]);
```
alexkli, line 575: As mentioned above, we still want the `timeout` metric, so should assert it in the tests here.

lib/worker.js
jdelbick: The reason I chose exit code 124 is because this is the exit code the [`timeout`](https://www.howtogeek.com/423286/how-to-use-the-timeout-command-on-linux/#:~:text=The%20exit%20code%20is%20124,code%20from%20timeout%20is%20zero.) function on Linux uses if it times out. It was slightly related to timing out so that is why I decided to use it. I am fine with changing to a different number. What would you suggest?

lib/worker.js
jdelbick: So the answer is currently not typically, because the timeout metric sends with 5s left and we start timing out with 15s left. I thought this error metric was supposed to replace the timeout metric. Instead of marking it as an `error` metric, we can define it as a `timeout` metric.

jdelbick:
@pheenomenon 

> (Same docker image is used in subsequent process calls.)
> 
> How do you verify this?

I verified this by setting up a timeout worker (firefly app) that essentially just sleeps for over a minute. I then ran `aio app run` and used the devtool to make a request. After the first request I went to new relic and got the name of the container the action ran in. then I ran a second request in the devtool and did the same thing. This proved in the different cases whether the action reused the container between requests.

> Also, I was wondering if this change also effects the multi-part rendition processing ? thoughts?
We determined the easiest solution for now is a hard kill after sending io events and metrics when the action is about to time out. So there is no difference in behavior whether it is in middle of normal processing or post-processing, or uploading, it will do a hard `process.exit()`

lib/worker.js
jdelbick: Ah nice catch, I had actually run into this edge case in manual testing. I had changed it but I guess I forgot to commit my changes.

jdelbick: I also found one more edge case in testing:

The `setTimeout` is defined in the [constructor](https://github.com/adobe/asset-compute-sdk/blob/ee89b351816af55d7640ebfae9d21df304fbd2ae/lib/worker.js#L77). And the send IO events, it relies on looking at the [`this.renditions`](https://github.com/adobe/asset-compute-sdk/blob/ee89b351816af55d7640ebfae9d21df304fbd2ae/lib/worker.js#L392) (which are all `Rendition` objects).  However, the `this.renditons` does not get defined until the [`prepare()`](https://github.com/adobe/asset-compute-sdk/blob/ee89b351816af55d7640ebfae9d21df304fbd2ae/lib/worker.js#L164) function, so if for some reason the code gets stuck and approaches time out before the `this.renditions` gets defined,  no events get sent. 

To handle this case, I went with the assumption that if `this.renditions` is not defined, we can assume no renditions have been successfully produced yet. This is because the code either errored or approached time out before it reached the point in the code where it defines the `this.renditions`. So, in this case, we can look at the `this.params.renditions` and send `rendition_failed` events for all of those renditions.

(btw I reached this edge case in testing by making a worker with a timeout = 15s so it would begin timing out immediately upon invoking the `worker()` function. I don't expect this case to come up too much in practice, but we should still handle it.)

lib/worker.js
jdelbick: renamed the `error` metric sent here to a `timeout` metric.

lib/worker.js
jdelbick: Default socket timeout is [30 seconds](https://github.com/adobe/node-fetch-retry#default-behavior) (and I confirmed we use the default value). However, this is [automatically adjusted](https://github.com/adobe/node-fetch-retry/blob/master/index.js#L57-L60) based on `retryMaxDuration` and `retryMaxDuration` is automatically adjusted for [openwhisk action timeout](https://github.com/adobe/node-fetch-retry#apache-openwhisk-action-support)...a bit confusing. 

AFAICS,
`retryMaxDuration` gets adjusted to action deadline if it is greater than action deadline. Then `socketTimeout` will get adjusted to half of the `retryMaxDuration`. So when we send IO events because the action will timeout in ~15s, if my math is correct, it will make the `retryMaxDuration` = ~15s and the the `socketTimeout` = ~7.5s which will result in about 2 retries (maybe more since we know the action won't actually timeout after 15 seconds)

alexkli: Ah, if that's the case, 2 retries sounds good!

alexkli:
@pheenomenon FWIW, we track the following attributes in [all metrics](https://github.com/adobe/node-openwhisk-newrelic#standard) which allow us to identify container reuse:

- `activationHost`
- `activationContainerName`

alexkli:

PR #22, opened 2020-07-16
NUI-668 Image rendition operations as post-processing in SDK

alexkli:
For [NUI-668](https://jira.corp.adobe.com/browse/NUI-668).

Work done mostly in mob session.


alexkli:
The one new test case added [fails](https://travis-ci.com/github/adobe/asset-compute-sdk/jobs/364049043) on travis since it does not have graphicsmagick installed, as expected (not listed [here](https://docs.travis-ci.com/user/reference/xenial/)).

See the log output of the buffer which you have to read top-down, it says `image conversion failed, cannot locate graphicsmagick`.

We'll have to see how we change that "end to end" test (maybe skip if gm is not present?) or if we can install gm inside `.travis.yml`. We have some precedence of installing extra software [here](https://github.com/adobe/asset-compute-devtool/blob/master/.travis.yml#L37).

tmathern:
TO-DO list for the SDK:
[ ] error handling for post-processing in worker.js
[ ] add post-processing to batch worker
[ ] fix timers
[ ] update SDK documentation

tmathern:
@alexkli Would you mind reviewing this a bit? In case you want some changes?

lib/postprocessing/image.js
```diff
@@ -12,56 +12,55 @@
...
+const gm = require("./gm-promisify");
+const util = require('util');
+const exec = util.promisify(require('child_process').exec);
+const fs = require('fs').promises;
 
-// Gets file size
-function fileSize(file) {
-    const stats = fs.statSync(file);
-    return stats.size;
+async function logImageMagickVersion() {
```
tmathern, line 20: Do we need that in prod?

lib/postprocessing/image.js
```diff
@@ -94,113 +93,50 @@ function processBasicImageOperations(img, instructions) {
...
-            img = img.resample(xdpi, ydpi);
+            img.resample(xdpi, ydpi);
         } else {
-            img = img.resample(instructions.convertToDpi, instructions.convertToDpi);
+            img.resample(instructions.convertToDpi, instructions.convertToDpi);
         }
     }
-
-    console.log('Processing rendition... Done!');
-    return img;
```
tmathern: I'd like you to keep that, so we can have a fluent API (aka you can chain function calls, since they return the `img` object).

lib/postprocessing/image.js
```diff
@@ -94,113 +93,50 @@ function processBasicImageOperations(img, instructions) {
...
-        // watermark embedded as base 64
-        console.log('Using embedded watermark');
-        watermark = 'inline:'+watermarkInstructions.watermarkContent;
-    }
-
-    img.dissolve(visibilityPercentage).gravity(placement);
-    img.composite(watermark);
+async function checkJpegSize(outfile, instructions) {
+    // If we specified jpegSize for JPEG output we need to verify that the size does not exceed that
+    if (instructions.fmt === "jpg" && instructions.jpegSize) {
```
tmathern, line 105: Use both jpg and jpeg. I some them both being used in worker tests - and fixing the test will time (you have to go look at them all).

lib/postprocessing/image.js
```diff
@@ -94,113 +93,50 @@ function processBasicImageOperations(img, instructions) {
...
-        img = processBasicImageOperations(img, instructions);
-    }
+            // From http://www.imagemagick.org/Usage/formats/#jpg
+            // As of IM v6.5.8-2 you can specify a maximum output filesize for the JPEG image.
+            img.define(`jpeg:extent=${instructions.jpegSize}`);
 
-    if(operations.watermarking) {
-        if(instructions.watermark && typeof instructions.watermark === 'object'){
-            img = processWatermark(img, instructions.watermark);
+            await img.write(outfile);
```
tmathern, line 115: Why are you not using the `write` function defined below?

lib/postprocessing/image.js
```diff
@@ -94,113 +93,50 @@ function processBasicImageOperations(img, instructions) {
...
         }
     }
+}
 
-    if(process.env.ASSET_COMPUTE_NO_METADATA_IN_IMG || operations.removeMetadata){
-        // removes profile and comments, for reproduceable images (for binary file comparision)
-        console.log('Removing profile and metadata from rendition');
-        img = img.strip();
-    }
+async function write(img, outfile, instructions) {
```
tmathern, line 120: Return `img` and it will become a fluent API...
(https://en.wikipedia.org/wiki/Fluent_interface)

I find it quite nice to have the postprocessing fluent. Feels more natural for image treatment.

lib/postprocessing/image.js
```diff
@@ -94,113 +93,50 @@ function processBasicImageOperations(img, instructions) {
...
-                                }
-                            });
-                    } else {
-                        console.log("END of rendition processing using image processing lib (imagemagick)", renditionName);
-                        resolve(true);
-                    }
-                }
-            });
-    });
+    const img = await read(infile);
```
tmathern, line 135: Not needed for now, but I would add an index as optional parameter to read.
This would be useful if a SDK user wants something else than the first picture from "something".

lib/postprocessing/image.js
```diff
@@ -94,113 +93,50 @@ function processBasicImageOperations(img, instructions) {
...
+
+    await render(img, instructions);
```
tmathern, line 137: Would also be nicer, I think, to be able to chain the calls (fluent API mentionned above).

lib/postprocessing/image.js
```diff
@@ -94,113 +93,50 @@ function processBasicImageOperations(img, instructions) {
...
+
+    await write(img, outfile, instructions);
```
tmathern, line 139: Would also be nicer, I think, to be able to chain the calls (fluent API mentionned above).

lib/worker.js
```diff
@@ -274,24 +259,12 @@ class AssetComputeWorker {
...
-                    console.error(`rendition post-processing for ${rendition.name} failed with error after ${postProcessingTimer.durationSec()}: ${err.message || err}`);
-                    this.timers.postProcessing += postProcessingTimer.duration(); // add post processing rendition time to total post processing duration
-                    await this.renditionFailure(rendition, err);
-                    continue;
-                }
-            } else {
-                console.log(`No postprocessing done for rendition ${rendition.name}`);
-                this.timers.postProcessing += postProcessingTimer.duration(); // add post processing rendition time to total post processing duration
+
+            rendition = await this.postProcess(rendition);
```
tmathern, line 263: `postProcess` does not return something in all cases (if postprocessing is successful, there is no object/value returned. So the condition below will evaluate to true, always). At least I didn't see a return statement?

lib/worker.js
```diff
@@ -208,29 +208,14 @@ class AssetComputeWorker {
...
-                console.log(`Postprocessing finished in ${postProcessingTimer.durationSec()} seconds: ${rendition.name}`);
-            } catch (err) {
-                console.error(`rendition post-processing failed with error after ${postProcessingTimer.durationSec()}: ${err.message || err}`);
-                this.timers.postProcessing += postProcessingTimer.duration(); // add post processing rendition time to total post processing duration
-                await this.renditionFailure(rendition, err);
-                return; // continue with next rendition
-            }
-        } else {
-            console.log(`No postprocessing done for rendition ${rendition.name}`);
+        rendition = await this.postProcess(rendition);
```
tmathern, line 215: `postProcess` does not return something in all cases (if postprocessing is successful, there is no object/value returned. So the condition below will evaluate to true, always). At least I didn't see a return statement?

test/rendition.postprocess.test.js
```diff
@@ -0,0 +1,682 @@
...
+ * governing permissions and limitations under the License.
+ */
+
+/* eslint-env mocha */
+/* eslint mocha/no-mocha-arrows: "off" */
+
+'use strict';
+
+const mockRequire = require("mock-require");
+const { MetricsTestHelper, GenericError } = mockRequire.reRequire("@adobe/asset-compute-commons");
```
jdelbick, line 19: remove unused `GenericError` to fix failing eslint in tests

lib/postprocessing/image.js
alexkli: It think it's very useful, because as we have seen, the imagemagick version is impossible to guess from the outside of an image (based on the OS) and we will always have some slight different images in prod, some workers might have a newer base version already, some not... Hence once we have a bug report about a particular rendition, we would know exactly the IM version right from the logs without having to do further forensics...

I am not sure about the cost of that `identify --version` call on every request. If we do that at startup of the container (i.e. upon loading of the module) we would not see it in the activation logs.

lib/postprocessing/image.js
alexkli: We do not need it at all and it allows to make `img` in functions that define it as local var (not as argument) a const. We aren't able to chain our own functions that way...

lib/postprocessing/image.js
alexkli: Good catch. This was in the existing code but seems we might have to adjust.

Aren't we normalizing in api-process or core? If not, shouldn't we do that there?

Related question: currently the output format is picked up by imagemagick purely because of the extension of the output file name, which we set to `fmt`. But there is a `gm().setFormat()` function. I feel we should make it more explicit. wdyt?

lib/postprocessing/image.js
alexkli: It might be hard to read in the diff: our `write()` function here calls the `checkJpegSize()` that this line of code is in. If we would reuse it there might be a recursion risk. There wouldn't be much reuse benefit since we only need the `img.write()` here.

lib/postprocessing/image.js
alexkli: See above.

lib/postprocessing/image.js
alexkli: Yeah, we have this `[0]` inside the `read()` there from the inherited code from worker-imagemagick. Not sure if it's actually used since we have dedicated workers for pdf and other document formats.

I think we can tackle it when we have more requirements, as it might also be an api addition that I wouldn't want to rush before we know exactly what we need.

lib/postprocessing/image.js
alexkli: Same comment as above:
1. for these 3 lines (read, render, write) there is no benefit in fluent. Fluent api is good if you have an API that is used _many_ times differently, setting various configs, for short pieces of code that represent specific configs/operations. But we don't have that here, since we have a generic bridge from our instructions API to the imagemagick/gm API. In such cases fluent doesn't bring anything to the table.
2. assuming we really wanted to have a fluent API, having it on `img` doens't help, we'd have to turn `render()` and `write()` into new classes with a fluent API, since these are own functions that are not part of `img` (`gm` API).

lib/worker.js
alexkli: It should and the tests forced me to get it right (spend some time on it). It returns undefined when there was a failure: https://github.com/adobe/asset-compute-sdk/blob/mob/NUI-668/lib/worker.js#L310

lib/worker.js
alexkli: See above.

lib/postprocessing/image.js
alexkli: The main reason I removed it was because it removed a considerable amount of `img = ` from the code all over the place, making it easier to read. Also it clears up that this isn't required and makes things easier since you eg. can have `img` as a `const` instead of a `let`, which will surely be helpful in future code additions here.

lib/postprocessing/image.js
tmathern: `Fluent api is good if you have an API that is used many times differently, setting various configs, for short pieces of code that represent specific configs/operations` <= that is what I expect to start happening when post-processing is in sdk. I think some people will look into our code and start calling this directly to do image processing. Hence my suggestion for the api to be nice and fluent, even if for now it is a private one. It won't hurt doing it - also what is the point of having `img` as const? If it is for performance gains they're minimal.

lib/postprocessing/image.js
tmathern: The indexing is to select the first image in multipage files - like pdf. And it was used in the pdf worker to select the first page of the pdf (that changed now in another PR), in order to only work on that first-page image.

lib/postprocessing/image.js
tmathern: I saw it. What I am saying is we could split that up a bit more to reuse write, and build a "wrapper" function calling the things in the proper order.

lib/postprocessing/image.js
tmathern: Well, I'd return `img` in all cases... If there is no operation done on `img`, the function would be a noop.

lib/postprocessing/image.js
tmathern: We are normalizing, but worker tests do not go through api-process (that is where the normalization happens), and they often still use old param names or mix up jpg and jpeg or tif and tiff. 

`But there is a gm().setFormat() function. I feel we should make it more explicit.` <= I think that too. Wasn't sure if you would agree to, so didn't ask before.

lib/postprocessing/image.js
tmathern: `the imagemagick version is impossible to guess from the outside of an image (based on the OS)` it should be in the metadata imagemagick writes (as I recall with don't remove them).

alexkli: ok. do we know if that is added to all output formats or just jpeg?

alexkli: Regarding the performance: on my Mac with imagemagick 7.0.10 printing the version takes ~20 msec +/- 10. That looks not too bad.

```
> time identify --version
Version: ImageMagick 7.0.10-28 Q16 x86_64 2020-08-16 https://imagemagick.org
Copyright: © 1999-2020 ImageMagick Studio LLC
License: https://imagemagick.org/script/license.php
Features: Cipher DPC HDRI Modules OpenMP(3.1) 
Delegates (built-in): bzlib freetype gslib heic jng jp2 jpeg lcms lqr ltdl lzma openexr png ps tiff webp xml zlib

real	0m0.021s
user	0m0.006s
sys	0m0.011s
```

One thing we could do is print this after the renditions have been uploaded so it is outside the "hot" code path, but still in the activation logs.

lib/postprocessing/image.js
alexkli: Actually we do not normalize anything wrt to `jpg` or `jpeg` in api-process. Can't find anything for

- [jpeg in api-process](https://git.corp.adobe.com/nui/api-process/search?utf8=✓&q=jpeg&type=)
- [jpeg in sdk](https://github.com/adobe/asset-compute-sdk/search?q=jpeg&unscoped_q=jpeg)

In the worker selection in `core` we always check both `jpg` and `jpeg`: https://git.corp.adobe.com/nui/core/blob/master/lib/workers.js#L236

Which seems to indicate we rely on workers to handle that themselves and in many cases rely on gm/im support both via the output extension... I think we could improve that and standardize on say `jpg` and normalize for that in api-process. Created [NUI-817](https://jira.corp.adobe.com/browse/NUI-817).

For now I will handle both in the post processing.

lib/postprocessing/image.js
alexkli: We don't offer the `gm` or any other javascript API to custom workers here. The only API is the rendition instructions.

If a custom worker wants to do something more complex with imagemagick (once they get access to it), they can use the `gm` library themselves in their code. But that would be separate from our (internal) post processing code.

lib/postprocessing/image.js
tmathern: I checked a png with cli exiftool and imagemagick's identify, it doesn't write into PNG metadata that it has been processed by imagemagick.

lib/postprocessing/image.js
alexkli: Got it - did that here: https://github.com/adobe/asset-compute-sdk/pull/22/commits/d44fe4af5d011f217d0e485ec63e8d55ef0d52a7#diff-c8cfb029f1e204a8f6da461ffa1464b0R103-R113

(now the command is logged for that possible 2nd execution as well)

alexkli: Ok, and I [removed](https://github.com/adobe/asset-compute-sdk/pull/22/commits/c5ec37f6a9f23a90191b4244ee6d740b28c9ab6c) the need for the 2nd execution, and the code became simpler (make sure to scroll down in that code snippet): https://github.com/adobe/asset-compute-sdk/blob/c5ec37f6a9f23a90191b4244ee6d740b28c9ab6c/lib/postprocessing/image.js#L110-L133

lib/rendition.js
```diff
@@ -104,10 +104,10 @@ async function detectContentType(renditionPath) {
  * Rendition abstraction
  */
 class Rendition {
-    constructor(instructions, directory, index = 0) {
+    constructor(instructions, directory, index = 0, name) {
```
tmathern, line 107: Usually you put default params last - or also give name a default value.

lib/shell/shellscript.js
```diff
@@ -74,6 +74,18 @@ class ShellScriptWorker {
             } else {
                 console.log('No content type information file found');
             }
+
+            if (await fs.pathExists(preparedFiles.optionsFile)) {
+                try {
+                    const options = await fs.readJSON(preparedFiles.optionsFile);
+                    if (options.postProcess === true || options.postProcess === "true") {
+                        rendition.postProcess = true;
```
tmathern, line 82:
```suggestion
                        rendition.postProcess = (options.postProcess === true || options.postProcess === "true");
```
And remove the `if`condition.

lib/postprocessing/image.js
```diff
@@ -30,6 +30,14 @@ async function read(infile) {
 }
 
 async function applyOperations(img, instructions) {
+    // handle exif orientation
+    // MUST happen before img.resize() for orientation 5 to 8 (otherwise width & height must be swapped)
+    // not using img.autoOrient() because that does -strip which strips the entire exif metadata
+    // but we only want to remove the exif Orientation flag which we simply unset
+    img.out("-auto-orient");
+    img.out("-orient");
+    img.out("Undefined");
```
tmathern, line 39: What is that for?

alexkli: See the comment above. `-orient Undefined` unsets/removes the orientation flag from the metadata.

alexkli: Each argument must be separate with `out()`. `img.out("-orient Undefined")` does not work.

alexkli:
As per the plan in [NUI-787](https://jira.corp.adobe.com/browse/NUI-787), we will merge this now (all tests are green, and the feature is opt in), and update workers incrementally. If there are fixes needed in the post processing logic for certain workers, we'll do them in new, focused PRs. Most work is expected to finish before next deployment of nui/app anyway - although it could also be feasible to deploy with post processing in just some workers (doesn't have to be advertised).

.gitignore
```diff
@@ -72,4 +72,6 @@ dist/
 !.github/**
 
 # any log
-*.log
\ No newline at end of file
+*.log
+
+.DS_Store
```
pheenomenon, line 77: also gitignore test-worker/test/asset-compute/worker/build ?

lib/postprocessing/image.js
```diff
@@ -0,0 +1,306 @@
...
+    console.log("skipping post-processing because intermediate rendition meets the target");
+    return false;
+}
+
+// ----------------------------- image conversion -----------------
+
+async function logImageMagickVersion() {
+    const { stdout } = await exec("identify --version");
+    console.log(".....................................................................................................");
+    console.log(stdout.toString().trim());
```
jdelbick, line 178: add imagemagick version as a metric maybe?

lib/postprocessing/gm-promisify.js
```diff
@@ -0,0 +1,38 @@
...
+    const img = GM(...arg);
+
+    // promisify gm output functions with callbacks
+    wrapAndPromisify(img, "write");
+    wrapAndPromisify(img, "stream");
+    wrapAndPromisify(img, "toBuffer");
+    wrapAndPromisify(img, "compare");
+    wrapAndPromisify(img, "thumb");
+    wrapAndPromisify(img, "format");
+    wrapAndPromisify(img, "identify");
```
pheenomenon, line 33: can we also add `wrapAndPromisify(img, "size");` to the promisified operations ?
pls see comment https://git.corp.adobe.com/nui/integration-tests/pull/77#discussion_r3075057

We could use `const GM = require('@adobe/asset-compute-sdk/lib/postprocessing/gm-promisify')` for reusability ?

lib/storage.js
```diff
@@ -22,21 +22,15 @@ const path = require('path');
...
-    return fs.existsSync(fullPath);
-}
-
 async function getSource(paramsSource, inDirectory, disableSourceDownload) {
     // Note: validation has happened in validate.js before
 
     if (process.env.WORKER_TEST_MODE) {
         // local file support for `Asset Compute test-worker` unit tests
         // not supported for clients in production
-        if (!isValidLocalFile(paramsSource.url, inDirectory)) {
```
pheenomenon, line 39: I fail to recollect, why we replaced `isValidLocalFile` with `!fs.existsSync(filePath)`

test-worker/.gitignore
```diff
@@ -0,0 +1 @@
+package-lock.json
```
pheenomenon, line 1: think we can remove this `.gitignore` file, and use the  `.gitignore` at root instead ?

lib/postprocessing/gm-promisify.js
alexkli: Will add it, but in a separate PR ok? Otherwise we have to wait until all the tests run again and we want to get this merged.

> We could use const GM = require('@adobe/asset-compute-sdk/lib/postprocessing/gm-promisify') for reusability ?

No, we should not do that. These deeper paths (`lib/`) really should stay an implementation detail of the module, so that we can change them without having to worry about breaking something.

As mentioned elsewhere, if we want to reuse this, we could have a small `gm-promisify` npm module for that.

pheenomenon:
I think good idea to add test-worker to `.npmignore`

lib/storage.js
alexkli: It's actually something I did just last week because of the `/in` and `/out` changes: instead of reading/writing directly to those folders that are mounted in the container (and which lead to problems), we copy instead. This also meant we had to adjust this validation logic and build the right path: `/nodejsAction/xyz/in/` plus `file.jpg`. The input params set by `test-worker` still have just the filename in them, e.g. `file.jpg`, so this made it easy.

See commit https://github.com/adobe/asset-compute-sdk/pull/22/commits/ea0ad47fb6de5e5fb3c5e31bbffada328cd7e267

test-worker/.gitignore
alexkli: No, we only want to gitignore the `package-lock.json` file here in the `test-worker/` directory (to make this test worker's npm always use latest versions), but not in the root (our sdk package.json).

alexkli:
@pheenomenon Good catch! This brings it down from 

```
npm notice package size:  25.8 MB                                 
npm notice unpacked size: 67.7 MB                                 
npm notice total files:   5868
```

to
```
npm notice package size:  29.5 kB                                 
npm notice unpacked size: 115.3 kB                                
npm notice total files:   24
```

lib/postprocessing/gm-promisify.js
alexkli: Turns out we need other changes anyway, so I will add it right away.

test-worker/.gitignore
pheenomenon: ok, so should we add `build` in this `.gitignore` ?

.gitignore
alexkli: that `build` directory is already gitignored.

alexkli: https://github.com/adobe/asset-compute-sdk/pull/22/files/a6b9fd5d3f889765aa36734a64fe4bacecc4e6ea#diff-a084b794bc0759e7a6b77810e01874f2R28

lib/postprocessing/image.js
alexkli: I think we don't need to look it up very often, and if we do, seeing it in splunk should be enough...

alexkli:
Only the flaky `worker-pdfrasterizer` test failed in `dependent_module_tests`: https://circle.ci.adobeitc.com/gh/nui/ci-opensource/7653

lib/storage.js
```diff
@@ -67,6 +61,14 @@ async function putRendition(rendition) {
         // asset-compute-cli command run-worker wants file named as originally requested
         // however, computing the metadata wants the current path, so
         // we need to make a copy not just rename
+        if(rendition.name.includes('post-')) {
+            // to verify post-processing renditions, since their names contain `post-` in the beginning
+            rendition.name = rendition.name.substring(5);
+
+            const newPath = path.join(rendition.directory, rendition.name);
+            await fs.copyFile(rendition.path, newPath);
```
alexkli, line 69: This might be expensive? Maybe rename/move is better.

PR #30, opened 2020-07-30
FEATURE-RELEASE: Embed small renditions in io events

jdelbick:
## Description

- based of POC in PR: https://github.com/adobe/asset-compute-sdk/pull/12
- default behavior is opt-out of embedded renditions
- to opt-in, you must set the `embedBinaryLimit` on the rendition object:
```js
{
source: "https://example.com/MySourceFile.jpg"
renditions: [{
        fmt: "png",
        embedBinaryLimit: 32 * 1024 // 32 kb is the max value
       }]
}
```
- io event `rendition_created` will store the rendition as a Data Uri in the `data` field of the event:
```js
{
      type: "rendition_created",
      rendition: {
      fmt: "png",
      embedBinaryLimit: embedBinaryLimit
      },
      data: 'data:application/octet-stream;base64,aGVsbG8gd29ybGQ=',
     source: "https://example.com/MySourceFile.jpg"
}
```
Notes:
- if you specify an `embedBinaryLimit` that is greater than our max limit (32kb), it will not embed the rendition (this is fragile...might be better, in this case, to set the `embedBinaryLimit` to be the max limit (32kb) so the rendition will still be embedded?)
- if the size of the rendition is greater than the defined `embedBinaryLimit`, it will not embed
- if the rendition file is bigger than the data uri limit of ~2MB and you set the `embedBinaryLimit` field, it will throw a `GenericError`
- the rendition data uri is not included in metrics

Fixes # https://jira.corp.adobe.com/browse/NUI-613

**still need to update the documentation & maybe an integration test covering this??**

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

## Checklist:

- [X] I have signed the [Adobe Open Source CLA](https://opensource.adobe.com/cla.html).
- [X] My code follows the code style of this project.
- [X] My change requires a change to the documentation.
- [X] I have read the **CONTRIBUTING** document.
- [X] I have added tests to cover my changes.
- [X] All new and existing tests passed.

### DIFF
--- a/lib/rendition.js
+++ b/lib/rendition.js
@@ -13,21 +13,24 @@
 'use strict';
 
 const util = require('util');
-
+const { GenericError } = require('@adobe/asset-compute-commons');
 const fs = require('fs-extra');
 const path = require('path');
 const crypto = require('crypto');
-
 const imageSize = util.promisify(require('image-size'));
 const contentType = require('content-type');
 const FileType = require('file-type');
-
 const exec = util.promisify(require('child_process').exec);
 
 const RENDITION_BASENAME = 'rendition';
 const DEFAULT_MIME_TYPE = 'application/octet-stream';
 
 const HASH_ALGORITHM = 'sha1';
+
+// 32 kb for inlined in IO events
+const EMBED_LIMIT_MAX = 32 * 1024;
+// roughly 2MB total based on https://stackoverflow.com/questions/695151/data-protocol-url-size-limitationsrenditions
+const DATA_URI_LIMIT = 2 * 1024 * 1024 - 100;
 
 const Metadata = {
     REPO_SIZE: 'repo:size',
@@ -277,6 +280,26 @@ class Rendition {
         // remove target URLs, could be sensitive
         delete obj.target;
         return obj;
+    }
+
+    embed() {
+        return Number.isInteger(this.instructions.embedBinaryLimit)
+            && this.instructions.embedBinaryLimit <= EMBED_LIMIT_MAX
+            && this.size() <= this.instructions.embedBinaryLimit;
+    }
+
+    asDataUri() {
+        // should never happen as embedBinaryLimit is limited, but extra safeguard against reading large files
+        if (this.size() > DATA_URI_LIMIT) {
+            throw new GenericError(`Rendition too large for data uri ${this.name}`);
+        }
+        const data = fs.readFileSync(this.path);
+
+        // cannot have spaces in data uri
+        // convert 'text/plain; charset=utf-8' into 'text/plain;charset=utf-8'
+        const type = this.contentType().replace(" ", "");
+
+        return `data:${type};base64,${data.toString("base64")}`;
     }
 
     static forEach(renditions, outDirectory) {
--- a/lib/rendition.js
+++ b/lib/rendition.js
@@ -288,7 +288,7 @@ class Rendition {
             && this.size() <= this.instructions.embedBinaryLimit;
     }
 
-    asDataUri() {
+    async asDataUri() {
         // should never happen as embedBinaryLimit is limited, but extra safeguard against reading large files
         if (this.size() > DATA_URI_LIMIT) {
             throw new GenericError(`Rendition too large for data uri ${this.name}`);
@@ -297,7 +297,7 @@ class Rendition {
 
         // cannot have spaces in data uri
         // convert 'text/plain; charset=utf-8' into 'text/plain;charset=utf-8'
-        const type = this.contentType().replace(" ", "");
+        const type = (await this.contentType()).replace(" ", "");
 
         return `data:${type};base64,${data.toString("base64")}`;
     }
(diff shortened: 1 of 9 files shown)


lib/worker.js
```diff
@@ -303,7 +304,7 @@ class AssetComputeWorker {
             uploadDuration: this.timers.currentUpload && this.timers.currentUpload.duration(),
             renditionDuration: durationSec(this.processingStartTime, renditionDoneTime),
             // rendition metadata
-            size: rendition.size()
+            size: rendition.size(),
```
alexkli, line 307: Nitpick: Comma not needed.

lib/rendition.js
```diff
@@ -279,6 +282,26 @@ class Rendition {
         return obj;
     }
 
+    embed() {
```
alexkli, line 285: Maybe naming it `shouldEmbed()` or `shouldEmbedInIOEvent()` might be more self-explaining.

tmathern: +1 for `shouldEmbedInIOEvent()`

lib/rendition.js
```diff
@@ -279,6 +282,26 @@ class Rendition {
...
+        return Number.isInteger(this.instructions.embedBinaryLimit)
+            && this.instructions.embedBinaryLimit <= EMBED_LIMIT_MAX
+            && this.size() <= this.instructions.embedBinaryLimit;
```
tmathern, line 288: Try to add a unit test which embeds something that has the size limit? Wdyt?

tmathern:
+1 also for adding an integration test. You can use one of the small images used for sdk tests. 
Also add a negative integration test case, to check error behavior.
But this ^ is going to be in another PR :) !

lib/rendition.js
tmathern: (You can generate a random array buffer with a given size - the unit test I mention doesn't really care about the data. I just want to have a test to verify the size limited is included and will upload properly!)

lib/rendition.js
```diff
@@ -279,6 +282,26 @@ class Rendition {
...
+    }
+
+    async asDataUri() {
+        // should never happen as embedBinaryLimit is limited, but extra safeguard against reading large files
+        if (this.size() > DATA_URI_LIMIT) {
+            throw new GenericError(`Rendition too large for data uri ${this.name}`);
+        }
```
pheenomenon, line 295: is this because asDataUri() is always expected to be invoked after embed() was called ? 
Ex:
 `data: rendition.embed() ? (await rendition.asDataUri()) : undefined`

tmathern: Data uris have a size limit of their own.

tmathern: (And that limit varies. So here we want to stay under DATA_URI_LIMIT. Take a look at https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs, the problems section for instance)

lib/rendition.js
jdelbick: I have a test where the `rendition.size()` > `this.instructions.embedBinaryLimit`: https://github.com/adobe/asset-compute-sdk/blob/d1cf3f23e7a00c11eca8927d45f4ba9b9dbd8321/test/api.test.js#L751 (by making the embedBinaryLimit super small)

and a test inside rendition tests where the `rendition.size()` > `this.instructions.embedBinaryLimit`: https://github.com/adobe/asset-compute-sdk/blob/d1cf3f23e7a00c11eca8927d45f4ba9b9dbd8321/test/rendition.test.js#L303 (by mocking the size of the rendition to be very large)

could you explain more what you mean by `something that has the size limit`? (size limit depends on user parameter `embedBinaryLimit` so my tests already cover this). I chose not to do a test where the rendition size is actually very large (maybe even the limit 32kb) because that would slow down testing

lib/rendition.js
jdelbick: yes, in theory this case should never be reached since `asDataUri()` is only used if `this.size()` < EMBED_LIMIT_MAX (32kb).

lib/rendition.js
tmathern: I'm looking for a rendition of size 32kb indeed. How come this would slow down testing? 32kb is still quite small... ?

pheenomenon:
@jdelbick  this would also need updates to API docs. Do you know if its handled in a separate ticket.

jdelbick:
@pheenomenon 
there is this ticket which is sort of related: https://jira.corp.adobe.com/browse/NUI-724
we will have to wait to document this in our api official docs until the feature is released in aem

lib/rendition.js
jdelbick: sorry I was confusing this with the data uri limit

alexkli:
@jdelbick FWIW: +1 for an integration test setting `embedBinaryLimit` with a very simple and fast request and verifying the result `data` property for the expected binary.

jdelbick:
@alexkli IT test already created in this PR: https://git.corp.adobe.com/nui/integration-tests/pull/63

PR #37, opened 2020-08-13
fix timers

jdelbick:
## Description
honestly, the timers are really a mess... it was a struggle to just get the timers correct without refactoring.

The issue is in batchWorker, we don't really have a `currentCallbackProcessing` and a `currentPostProcessing` because they happen in batch. But also, we technically do have a `currentPostProcessing` since we post-process one by one. So it leaves a weird position, in rendition metrics for batch workers, the `callbackProcessDuration` metric is equal to the total callbackProcessDuration for all the renditions. So following this, it would make the most sense to make the `postProcessDuration` metric for each rendition the total `postProcessDuration`....an easy way to do this would be simply setting the `this.timers.currentPostProcessing` timer equal to the `this.timers.postProcessing` timer...except that `this.timers.postProcessing` is __not__ a timer but a duration. Very confusing. 

I wanted to keep the postProcess code the same in batchWorker() and worker() so the `this.timers.currentPostProcessing` resets with each rendition inside batchWorker post processing. Side effect of this means that `postProcessDuration` metric for each rendition is the value of the __last rendition__  `this.timers.currentPostProcessing.duration()`. See my notes in the test.

When we refactor we need to make all of the following a timer or a duration:
```
this.timers = {
            actionDuration: new Timer(), // timer
            download: 0, // timer??
            processing: 0, // duration
            postProcessing: 0, // duration
            callbackProcessing: 0, // duration
            upload: 0, // timer
            currentProcessing: 0, // timer
            currentPostProcessing: 0, // timer
            currentUpload: 0 // timer
        };
```

Fixes # https://jira.corp.adobe.com/browse/NUI-766

## Types of changes

## Checklist:

### DIFF
--- a/lib/worker.js
+++ b/lib/worker.js
@@ -214,12 +214,12 @@ class AssetComputeWorker {
         }
 
         // 2. check and log resulting rendition
+        this.timers.callbackProcessing += callbackTimer.end();
         console.log(`worker() callback generated rendition in ${callbackTimer.durationSec()} seconds: ${rendition.name}`);
 
         // 3. post processing for images
-        this.timers.currentPostProcessing = new Timer();
-        const postProcessingTimer = this.timers.currentPostProcessing;
-
+        // start post-processing timer
+        const postProcessingTimer = this.timers.currentPostProcessing = new Timer();
         const shouldDoPostProcess = await this.shouldPostProcess(rendition);
         if (shouldDoPostProcess) {
             try {
@@ -239,7 +239,6 @@ class AssetComputeWorker {
                 rendition = postProcessedRendition;
                 this.renditions[rendition.index] = postProcessedRendition;
 
-                postProcessingTimer.end();
                 console.log(`Postprocessing finished in ${postProcessingTimer.durationSec()} seconds: ${rendition.name}`);
             } catch (err) {
                 console.error(`rendition post-processing failed with error after ${postProcessingTimer.durationSec()}: ${err.message || err}`);
@@ -251,7 +250,6 @@ class AssetComputeWorker {
         }
 
         this.timers.postProcessing += postProcessingTimer.duration();
-        this.timers.callbackProcessing += callbackTimer.end();
 
         if (this.options.disableRenditionUpload) {
             await this.renditionSuccess(rendition);
@@ -261,8 +259,7 @@ class AssetComputeWorker {
     }
 
     async batchProcessRenditions(renditionsCallback) {
-        const timer = this.timers.currentCallbackProcessing = new Timer();
-        //TODO: Check timers 
+        const callbackTimer = this.timers.currentCallbackProcessing = new Timer();
 
         // rendition callback execution
         try {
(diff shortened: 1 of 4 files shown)


lib/worker.js
```diff
@@ -274,16 +271,16 @@ class AssetComputeWorker {
             // call client-provided callback to transform source into 1 rendition
             await renditionsCallback(this.source, this.renditions, this.directories.out, this.params);
 
-            timer.end();
-            console.log(`processing finished successfully after ${timer.durationSec()} seconds`);
+            this.timers.callbackProcessing = callbackTimer.end();
```
jdelbick, line 274: btw, `timer.end()`, `.durationSec()`, and `.duration()` all end the timer so I could remove a lot of lines

test/rendition.postprocess.test.js
```diff
@@ -25,9 +25,9 @@ const fs = require('fs-extra');
 const { MetricsTestHelper } = require("@adobe/asset-compute-commons");
 
 const PNG_FILE = "test/files/fileSmall.png";
-const BMP_FILE = "test/files/file.bmp";
-const GIF_FILE = "test/files/file.gif";
-const JPG_FILE = "test/files/file.jpg";
+// const BMP_FILE = "test/files/file.bmp";
```
tmathern, line 28: Instead of commenting, why not remove it?

jdelbick: These were going to be used in the tests when the test is completed. for now I commented it out to ignore the eslint changes

tmathern:
@jdelbick Also create a ticket for the refactoring of timers, otherwise I fear it will be forgotten.

lib/worker.js
```diff
@@ -214,12 +214,12 @@ class AssetComputeWorker {
...
         // 2. check and log resulting rendition
+        this.timers.callbackProcessing += callbackTimer.end();
         console.log(`worker() callback generated rendition in ${callbackTimer.durationSec()} seconds: ${rendition.name}`);
 
         // 3. post processing for images
-        this.timers.currentPostProcessing = new Timer();
-        const postProcessingTimer = this.timers.currentPostProcessing;
-
+        // start post-processing timer
+        const postProcessingTimer = this.timers.currentPostProcessing = new Timer();
```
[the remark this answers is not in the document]
jdelbick: I actually found it more confusing when they were on separate lines. 

This is because when a variable is defined, I expect to use it. However, we don't expect to use `this.timers.currentPostProcessing` but instead, refer to it as `postProcessingTimer`. When they were defined on separate lines, I kept getting confused about whether to use `this.timers.currentPostProcessing` or `postProcessingTimer`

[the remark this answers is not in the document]
tmathern: Some people may read this different ways, let me explain: 
1. `const postProcessingTimer = this.timers.currentPostProcessing = new Timer();` can be understood as: create a new timer and assign it to `this.timers.currentPostProcessing`. Then, assign `this.timers.currentPostProcessing` to `postProcessingTimer`. This is also what the code will do here.
2. `const postProcessingTimer = this.timers.currentPostProcessing = new Timer();` can be understood as: try to instantiate a new Timer() and assign the object's reference to `this.timers.currentPostProcessing`. Then, if the assignment worker, assign 0 (true) to `postProcessingTimer`, or any other value otherwise. That is a behavior you will see in C code, as this would be what C would do (Note: this is actually a way in C to check if memory allocation worked or not, it's not some rare thing). Therefore, that line of code may confuse people coming from C/C++ background. (That's a reason why you I always tried to avoid that, because different people with different development background can be confused by that).

lib/worker.js
```diff
@@ -296,6 +293,8 @@ class AssetComputeWorker {
 
         // post-processing // TODO: Refactor this to share partas with post-processing in `processRendition` func
         for (let rendition of this.renditions) {
+            // start post-processing timer
+            const postProcessingTimer = this.timers.currentPostProcessing = new Timer();
```
pheenomenon, line 297: the only reason you are doing `postProcessingTimer` inside for(..) is because we care for the `postProcessingTimer.durationSec()` in logs correct? otherwise Metrics wise.. this can be done outside for..loop.
```
const postProcessingTimer = this.timers.currentPostProcessing = new Timer();

this.timers.postProcessing += postProcessingTimer.duration();
```

jdelbick: the reason I did this in the loop was for two reasons:
1. I wanted to keep the post-processing functions the same in `worker` and `batchWorker` since we plan to extract this function out and use it in both cases
2. Yes the timestamp is needed for logging purposes, not metrics as we don't track metrics one by one for each rendition in batchWorker

lib/worker.js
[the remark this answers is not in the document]
jdelbick: For now, I'll keep it as is (1 liner) since that's how we define it in other places in the code.

PR #39, opened 2020-08-14
add test for batch postprocessing

pheenomenon:
## Description

Adding unit tests for batch post processing

Fixes # 
https://jira.corp.adobe.com/browse/NUI-760

## Types of changes

- [x] no-release

## Checklist:


test/rendition.postprocess.test.js
```diff
@@ -227,18 +226,60 @@ describe("imagePostProcess", () => {
...
+            },],
+            requestId: "test-request-id",
+            auth: testUtil.PARAMS_AUTH,
+            newRelicEventsURL: MetricsTestHelper.MOCK_URL,
+            newRelicApiKey: MetricsTestHelper.MOCK_API_KEY
+        };
+        const result = await main(params);
+
+        // should fail
+        assert.ok(result.renditionErrors);
```
jdelbick, line 275: maybe check the error thats thrown here

test/rendition.postprocess.test.js
```diff
@@ -227,18 +226,60 @@ describe("imagePostProcess", () => {
...
+        assert.equal(events[2].rendition.fmt, "tiff");
+
+        const uploadedFileBase64_png = Buffer.from(uploadedRenditions["/MyRendition1.png"]).toString('base64');
+        const uploadedFileBase64_jpg = Buffer.from(uploadedRenditions["/MyRendition2.jpeg"]).toString('base64');
+        const uploadedFileBase64_tiff = Buffer.from(uploadedRenditions["/MyRendition3.tiff"]).toString('base64');
+        assert.ok(BASE64_RENDITION_JPG  === uploadedFileBase64_jpg);
+        assert.ok(BASE64_RENDITION_PNG  === uploadedFileBase64_png);        
+        assert.ok(BASE64_RENDITION_TIFF  === uploadedFileBase64_tiff);
+    });
+    it('should fail all worker batch renditions if postprocessing for one rendition fails', async () => {
```
jdelbick, line 238: looking at the logs from this test, it fails in rendition_upload:

```
Error: { GenericError: PUT 'https://example.com/MyRendition3.tiff' failed with status 404: <!doctype html>
<html>
<head>
    <title>Example Domain</title>

    <meta charset="utf-8" />
    <meta http-equiv="Content-type" content="text/html; charset=utf-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1" />
    <style type="text/css">
    body {
        background-color: #f0f0f2;
        margin: 0;
        padding: 0;
        font-family: -apple-system, system-ui, BlinkMacSystemFont, "Segoe UI", "Open Sans", "Helvetica Neue", Helvetica, Arial, sans-serif;
        
    }
    div {
        width: 600px;
        margin: 5em auto;
        padding: 2em;
        background-color: #fdfdff;
        border-radius: 0.5em;
        box-shadow: 2px 3px 7px 2px rgba(0,0,0,0.02);
    }
    a:link, a:visited {
        color: #38488f;
        text-decoration: none;
    }
    @media (max-width: 700px) {
        div {
            margin: 0 auto;
            width: auto;
        }
    }
    </style>    
</head>

<body>
<div>
    <h1>Example Domain</h1>
    <p>This domain is for use in illustrative examples in documents. You may use this
    domain in literature without prior coordination or asking for permission.</p>
    <p><a href="https://www.iana.org/domains/example">More information...</a></p>
</div>
</body>
</html>

    at Object.upload (/Users/delbick/work/jdelbick/asset-compute-sdk/lib/storage/http.js:79:19)
    at process._tickCallback (internal/process/next_tick.js:68:7)
  name: 'GenericError',
  location: 'test_action_upload',
  date: 1597443268887 }
```

this does not really test postprocessing failing

jdelbick: I am currently working on [failure tests for image post-processing in the sdk](https://jira.corp.adobe.com/browse/NUI-763)...a good solution is to mock `asset-compute-image-processing` (which we want to do for all of our tests anyways)

jdelbick: the other option / question is whether we should throw an error here: https://github.com/adobe/asset-compute-sdk/blob/NUI-760/lib/worker.js#L357 when we know we cannot do post-processing

pheenomenon: yup that's right, this `should fail all worker batch renditions if postprocessing for one rendition fails` sneaked in for something I wanted for mock `asset-compute-image-processing`. otherwise cannot trick the failure scenarios. I will take this test out from this PR & just merge the first onw -> `should download source, invoke worker in batch callback and upload rendition - different rendition`.

pheenomenon: @jdelbick Let me know on https://jira.corp.adobe.com/browse/NUI-763, it might intersect with https://jira.corp.adobe.com/browse/NUI-766

tmathern: The `shouldPostProcess ` should not throw, I think. This is a function to determine if we have to post-process, not if we can (and I'd like to use this function in if conditions). It's the difference between "having to" and "being able to". The throwing should happen later if we fail post-processing. 

The question is rather if a worker requests post-processing but we can't (`shouldPostProcess` returns false). Currently, we ignore the worker's request if we can't post-process (we don't post-process). The thing I could see coming is that we have another flag just for the watermark, which we could quite easily support since the watermark and image operations are independent in the lib (but the API is not public).

pheenomenon: for [batchProcessRenditions](https://github.com/adobe/asset-compute-sdk/blob/NUI-760/lib/worker.js#L299)  we would have to think about else for `shouldPostProcess`. 
we take the [all-fail approach for batch](https://github.com/adobe/asset-compute-sdk/blob/NUI-760/lib/worker.js#L287) for the scenario when `worker-pie` erred. So to support all-or-none in post-process we need to handle this scenario differently.

jdelbick: good point @pheenomenon, to be consistent we would have to take the all-or-none approach

tmathern: No. Having one rendition not having the same post-processing flag as the others should not make fail everything.
A failure in our case is basically if an error gets thrown.

PR #40, opened 2020-08-17
Nui 761 refactor postprocess for batchCallBack

pheenomenon:
## Description

https://jira.corp.adobe.com/browse/NUI-761 Refactor post-processing
Refactored ImagepostProcess function to be reused for single and batch post process.
Following up from https://github.com/adobe/asset-compute-sdk/pull/39#discussion_r470896671
>for [batchProcessRenditions](https://github.com/adobe/asset-compute-sdk/blob/NUI-760/lib/worker.js#L299)  we would have to think about else for `shouldPostProcess`. 
we take the [all-fail approach for batch](https://github.com/adobe/asset-compute-sdk/blob/NUI-760/lib/worker.js#L287) for the scenario when `worker-pie` erred. So to support all-or-none in post-process we need to handle this scenario differently.

The current design assumption(which is in line with renditionCallBack) is that if `shouldPostProcess` returns false for one rendition then fail postProcess for the batch.
Also changed error-handling as previous code was not sending out metrics as it was skipped.

Fixes # (issue)
https://jira.corp.adobe.com/browse/NUI-761

## Types of changes

## Checklist:


lib/worker.js
```diff
@@ -292,40 +292,31 @@ class AssetComputeWorker {
         }
 
         // post-processing // TODO: Refactor this to share partas with post-processing in `processRendition` func
+        // eslint-disable-next-line prefer-const
         for (let rendition of this.renditions) {
-            // start post-processing timer
-            const postProcessingTimer = this.timers.currentPostProcessing = new Timer();
```
tmathern, line 297: I think we still need to keep the timer's reference `this.timers.currentPostProcessing` somewhere - wdyt @jdelbick ?

lib/worker.js
```diff
@@ -292,40 +292,31 @@ class AssetComputeWorker {
         }
 
         // post-processing // TODO: Refactor this to share partas with post-processing in `processRendition` func
+        // eslint-disable-next-line prefer-const
```
tmathern, line 295: It should be OK to use `const rendition`, I don't think we reassign it.

lib/worker.js
```diff
@@ -341,6 +332,38 @@ class AssetComputeWorker {
...
+        const postProcessingEngine = this.postProcessingEngine || imgProcessingEngine;
+        await postProcessingEngine.imageProcess(rendition.path, postProcessedRendition.path, rendition.instructions);
+
+        // point to proper rendition once postprocessing done
+        rendition = postProcessedRendition;
+        this.renditions[rendition.index] = postProcessedRendition;
+    }
+
+    async handleBatchPostProcessFailure(postProcessingTimer,err) {
+        console.error(`rendition post-processing for batch failed after ${postProcessingTimer.durationSec()}: ${err.message || err}`);
```
tmathern, line 354: If we keep the timer's reference, you can get rid of the timer parameter in the call and use `this.postProcessingTimer` instead.

lib/worker.js
pheenomenon: I had to change the usage back to `let rendition` as inside `doPostProcess` we actually modify `rendition` so not returning the updated code seems deceiving.

lib/worker.js
pheenomenon: for now have brought back `this.timers.currentPostProcessing` but I think this is not accurate as is. For batch  currentPostProcessing = postProcessing so am leaving a TODO for now. (which I believe is handled in a separate ticket)

lib/worker.js
tmathern: You mean because the reference changes at this line: https://github.com/adobe/asset-compute-sdk/blob/3463b14a66469f4e97f2a1cfc5816f978e8cd31d/lib/worker.js#L349 ?

lib/worker.js
tmathern: @jdelbick dealt with the current timers implementation in depth recently. Could you take a look, @jdelbick ?

jdelbick: +1 it should be referenced with `this` instead of passing through a variable

lib/worker.js
pheenomenon: as we update the `rendition` here, for the new `rendition` to take effect after the method returns, it required to be reassigned https://github.com/adobe/asset-compute-sdk/blob/269770a37a8f9d1a65ea35bf96216d40997330bb/lib/worker.js#L226.

tmathern: So we need the new rendition only for the log statement? Because in that case you could get the name before the call in a variable. So you could remove the need to return a rendition too.

pheenomenon: actually we also needed changed `rendition` for https://github.com/adobe/asset-compute-sdk/blob/269770a37a8f9d1a65ea35bf96216d40997330bb/lib/worker.js#L242
which otherwise will use `rendition0.jpg` instead of `post-rendition0.jpg`

pheenomenon:
merging for now as its needed for other branches/tickets. happy to answer any more questions here thought.

PR #42, opened 2020-08-18
NUI-763 Image processing error tests and fixes

pheenomenon:
## Description

This is part of SDK Image post processing [ticket](https://jira.corp.adobe.com/browse/NUI-668). Adds tests to find and fix issues from mob-session
Changes to worker.js to fix:
- Timer inconsistency in single worker post-processing.
- Handle & fix metrics timers for batch-worker post-processing.
- Handle and fix metrics for batch-worker post-processing partial/all error
- Fix flow for batch-worker post-processing if only partial renditions are post-processing eligible
- Added tests for various combinations of batch-worker post-process failures

Fixes # (issue)
https://jira.corp.adobe.com/browse/NUI-763

## Types of changes

## Checklist:


lib/worker.js
```diff
@@ -25,6 +25,7 @@ const { validateParameters } = require('./validate');
 const Timer = require('./utils/timer');
 const { Sampler } = require('@adobe/metrics-sampler');
 const { metrics: cgroupMetrics, cpu } = require('@adobe/cgroup-metrics');
+const { fail } = require('assert');
```
jdelbick, line 28: was this imported on purpose?

test/rendition.postprocess.test.js
```diff
@@ -43,7 +43,7 @@ mockRequire("@adobe/asset-compute-image-processing", {
             } else if (instructions.fmt === 'tiff') {
                 await fs.copyFile('test/files/generatedFileSmall.tiff',outfile);
             } else {
-                throw new Error('unknown error');
+                throw new GenericError('batch worker renditions postprocess applies only for all renditions');
```
jdelbick, line 46: if you look at asset-compute-image-processing, we throw actual Errors (not GenericErrors): https://github.com/adobe/asset-compute-image-processing/blob/master/lib/image.js#L164 and that's why I had it throw an Error.

this error is then caught in worker.js and turned into a `GenericError`

lib/worker.js
```diff
@@ -274,34 +275,41 @@ class AssetComputeWorker {
             }
             return;
         }
-
+        // batch post-process is all or none. so using currentPostProcessing(is timer) as postProcessing(is duration)
+        this.timers.currentPostProcessing = new Timer();
+        let ppRendition = 0;
```
jdelbick, line 280: this variable name is confusing. IIUC, it is meant to track the amount of renditions that got post-processed?
maybe call it `postProccessedCount`

lib/worker.js
```diff
@@ -274,34 +275,41 @@ class AssetComputeWorker {
...
                 }
             } else {
-                const err = new GenericError('batch worker renditions postprocess applies only for all renditions');
+                console.log(`No postprocessing done for rendition ${rendition.name}`);
+            }
+            // by convention if a even a single rendition is not postprocess eligible then failAll.
+            // following block is only for reqeust with multiple rendition and if few are postprocess ineligible
+            // skip: if only 1 rendition and is not post process eligible
+            // fail fast: if more than one rendition and one or more failed
+            if(ppRendition > 0 && !shouldDoPostProcess) {
```
jdelbick, line 303: I am still confused about this. this is in case there has been at least one post-processed asset and another asset in the batch is not supposed to be post-processed.

1. we then can make `ppRendition` into a boolean instead of a counter. we only need to track whether at least 1 rendition was post processed. in this case lets make it a boolean and call it `postProcessed`

2. in the case where one of the renditions in the batch is post-processed and another is not post-processed, I would argue we don't need to fail here. we should fail the entire batch if one of the renditions fails during post-processing. but if one rendition needs post-processing and others don't I do not think we should fail the entire renditions.

lib/worker.js
```diff
@@ -337,10 +345,16 @@ class AssetComputeWorker {
...
     async handleBatchPostProcessFailure(err) {
         console.error(`rendition post-processing for batch failed after ${this.timers.currentPostProcessing.durationSec()}: ${err.message || err}`);
+        //for activation metrics
+        this.timers.postProcessing = this.timers.currentPostProcessing.duration();
         await this.metrics.handleError(err, {
             location: `${this.actionName}_batchProcess`,
             metrics: {
-                postProcessingDuration: this.timers.postProcessing + this.timers.currentPostProcessing.duration(),
+                callbackProcessingDuration: this.timers.currentCallbackProcessing ? this.timers.currentCallbackProcessing.duration() : 0,
+                // postProcessingDuration: this.timers.currentPostProcessing ? this.timers.currentPostProcessing.duration() : 0,
```
jdelbick, line 354: remove comment

test/rendition.postprocess.test.js
```diff
@@ -190,11 +193,19 @@ describe("imagePostProcess", () => {
...
         assert.equal(receivedMetrics[0].callbackProcessingDuration + receivedMetrics[0].postProcessingDuration, receivedMetrics[0].processingDuration);
         assert.equal(receivedMetrics[3].eventType, "activation");
         assert.equal(receivedMetrics[3].callbackProcessingDuration, receivedMetrics[0].callbackProcessingDuration, receivedMetrics[1].callbackProcessingDuration, receivedMetrics[2].callbackProcessingDuration);
         assert.equal(receivedMetrics[3].callbackProcessingDuration + receivedMetrics[3].postProcessingDuration, receivedMetrics[3].processingDuration);
+        // downloadDuration: 0.002481363,
+        // callbackProcessingDuration: 0.002246784,
+        // postProcessingDuration_id_mean: 449890804.5,
+        // postProcessingDuration_hrtimeElapsed_mean: 1115159,
+        // processingDuration: '0.002246784[object Object]',
+        // uploadDuration: 0.057330613999999995
```
jdelbick, line 207: remove comment

lib/worker.js
tmathern: (+1, please avoid acronyms in variable names).

lib/worker.js
```diff
@@ -274,34 +275,41 @@ class AssetComputeWorker {
...
             const shouldDoPostProcess = await this.shouldPostProcess(rendition);
             if (shouldDoPostProcess) {
                 try {
                     rendition = await this.doPostProcess(rendition);
                     console.log(`Batch Postprocessing finished in ${postProcessingTimer.durationSec()} seconds: ${rendition.name}`);
+                    ppRendition++;
                 } catch (err) {
-                    console.error(`rendition post-processing for batch failed with error after ${postProcessingTimer.durationSec()}: ${err.message || err}`);
+                    console.error(`rendition post-processing for ${rendition.name} failed with error after ${postProcessingTimer.durationSec()}: ${err.message || err}`);
+                    console.error(`rendition post-processing for batch failed after ${this.timers.currentPostProcessing.durationSec()}`);
```
tmathern, line 292: Could we merge the two console log statements into one?

lib/worker.js
pheenomenon: Not sure, it was previously required but not used anywhere and eslint was complaining.

lib/worker.js
tmathern: I think the logic is wrong, here. A rendition that does not need post-processing will also have shouldDoPostProcess return false (remember, post-processing needs a flag set). As far as I understand the describing comment, a rendition not requesting post-processing would also make the batch fail here... Which is not what we want, as some renditions may indeed not need post-processing.

tmathern: (Remember that the post-processing requesting flag is per rendition, not per batch).

lib/worker.js
```diff
@@ -274,34 +275,41 @@ class AssetComputeWorker {
...
-                const err = new GenericError('batch worker renditions postprocess applies only for all renditions');
+                console.log(`No postprocessing done for rendition ${rendition.name}`);
+            }
+            // by convention if a even a single rendition is not postprocess eligible then failAll.
+            // following block is only for reqeust with multiple rendition and if few are postprocess ineligible
+            // skip: if only 1 rendition and is not post process eligible
+            // fail fast: if more than one rendition and one or more failed
+            if(ppRendition > 0 && !shouldDoPostProcess) {
+                console.error(`rendition post-processing for batch failed after ${this.timers.currentPostProcessing.durationSec()}`);
+                const err = new GenericError('All rendition in batch worker must be postprocess eligible');
```
tmathern, line 305: I don't think this must be the case.

test/rendition.postprocess.test.js
tmathern: I would do fuzzy checks - I didn't see time being mocked and we can't exclude small variations. Or did I miss something in the code?

test/rendition.postprocess.test.js
```diff
@@ -258,4 +269,357 @@ describe("imagePostProcess", () => {
...
+        await MetricsTestHelper.metricsDone();
+        console.log('receivedMetrics--->', receivedMetrics);
+        assert.equal(receivedMetrics[0].eventType, "error");
+        assert.equal(receivedMetrics[0].callbackProcessingDuration + receivedMetrics[0].postProcessingDuration, receivedMetrics[0].processingDuration);
+        assert.equal(receivedMetrics[1].eventType, "activation");
+        assert.equal(receivedMetrics[1].callbackProcessingDuration, receivedMetrics[0].callbackProcessingDuration);
+        assert.equal(receivedMetrics[1].postProcessingDuration, receivedMetrics[0].postProcessingDuration);
+        assert.equal(receivedMetrics[1].processingDuration, receivedMetrics[0].processingDuration);
+    });
+    it('should fail all if some renditions are post process ineligible', async () => {
```
tmathern, line 372: Not sure about that. See review comments.

test/rendition.postprocess.test.js
```diff
@@ -258,4 +269,357 @@ describe("imagePostProcess", () => {
...
+        // check metrics
+        await MetricsTestHelper.metricsDone();
+        assert.equal(receivedMetrics[0].eventType, "error");
+        assert.equal(receivedMetrics[0].callbackProcessingDuration + receivedMetrics[0].postProcessingDuration, receivedMetrics[0].processingDuration);
+        assert.equal(receivedMetrics[1].eventType, "activation");
+        assert.equal(receivedMetrics[1].callbackProcessingDuration, receivedMetrics[0].callbackProcessingDuration);
+        assert.equal(receivedMetrics[1].postProcessingDuration, receivedMetrics[0].postProcessingDuration);
+        assert.equal(receivedMetrics[1].processingDuration, receivedMetrics[0].processingDuration);
+    });
+    it('should fail all if edge renditions are post process ineligible', async () => {
```
tmathern, line 432: What are "edge renditions" ?

test/rendition.postprocess.test.js
pheenomenon: thanks, had changed it for testing GenericError else where. will revert.

lib/worker.js
pheenomenon: yes this tracks the number of renditions. `postProccessedCount` it is.

lib/worker.js
jdelbick: I would remove it, should not be necessary in this file

lib/worker.js
pheenomenon: @jdelbick @tmathern 
Implementation to change after discussion on SLACK: https://cq-dev.slack.com/archives/GAMJLHSBH/p1597871482310900?thread_ts=1597870229.284100&cid=GAMJLHSBH

@jdelbick I remember trying the Boolean logic which I think failed existing tests.

pheenomenon:
closing this PR, to reduce the clutter. will open a new one.

PR #47, opened 2020-08-25
add option.json for shellscript opt-in post processing

jdelbick:
## Description

stores it in `out/options.json`:
```
{
   "postProcess": true
}
```
- must be json (but gracefully handles without failing if its not a json)
- must be boolean `true` and not `"true"` string

Fixes # https://jira.corp.adobe.com/browse/NUI-779

## Types of changes

## Checklist:

### DIFF
--- a/lib/shell/shellscript.js
+++ b/lib/shell/shellscript.js
@@ -79,9 +79,14 @@ class ShellScriptWorker {
             }
 
             if (await fs.pathExists(optionsFile)) {
-                const options = await fs.readJSON(optionsFile);
-                if (options.postProcess === true) {
-                    rendition.postProcess = true;
+                try {
+                    const options = await fs.readJSON(optionsFile);
+                    if (options.postProcess === true) {
+                        rendition.postProcess = true;
+                    }
+                // eslint-disable-next-line no-unused-vars
+                } catch (e) {
+                    // ignore error if options file is not properly formatted
                 }
             }
         } catch (err) {
--- a/lib/shell/shellscript.js
+++ b/lib/shell/shellscript.js
@@ -50,11 +50,8 @@ class ShellScriptWorker {
 
         const preparedFiles = await prepareMetadata(rendition.directory);
 
-        // rendition options file for optiosn like postProcess
-        const optionsFile = path.resolve(rendition.directory, 'options.json');
-
         // inherit environment variables
-        const env = setVariables(rendition, source, preparedFiles.errorFile, preparedFiles.typeFile, optionsFile);
+        const env = setVariables(rendition, source, preparedFiles);
 
         try {
             // use of spawn() with separate arguments prevents shell/command injection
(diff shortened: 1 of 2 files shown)


lib/shell/shellscript.js
```diff
@@ -50,8 +50,11 @@ class ShellScriptWorker {
 
         const preparedFiles = await prepareMetadata(rendition.directory);
 
+        // rendition options file for optiosn like postProcess
+        const optionsFile = path.resolve(rendition.directory, 'options.json');
+
         // inherit environment variables
-        const env = setVariables(rendition, source, preparedFiles.errorFile, preparedFiles.typeFile);
+        const env = setVariables(rendition, source, preparedFiles.errorFile, preparedFiles.typeFile, optionsFile);
```
tmathern, line 57: At that point, I would consider putting the various files to create in an object, to avoid keeping changing the `setVariables` signature.

lib/shell/shellscript.js
```diff
@@ -74,6 +77,18 @@ class ShellScriptWorker {
             } else {
                 console.log('No content type information file found');
             }
+
+            if (await fs.pathExists(optionsFile)) {
+                try {
+                    const options = await fs.readJSON(optionsFile);
+                    if (options.postProcess === true) {
```
tmathern, line 84: ptions.postProcess === true || ptions.postProcess === 'true' 

In case the boolean ever gest parsed as string.

lib/shell/shellscript.js
jdelbick: yeah, well hopefully this `optionsFile` will cover any remaining things we would write to files...so no more variables

lib/shell/shellscript.js
```diff
@@ -50,8 +50,11 @@ class ShellScriptWorker {
 
         const preparedFiles = await prepareMetadata(rendition.directory);
 
+        // rendition options file for optiosn like postProcess
+        const optionsFile = path.resolve(rendition.directory, 'options.json');
```
pheenomenon, line 54: I think `prepareMetadata` is doing something similar. Have you considered encapsulating `options.json` inside it i.e
```
    return {
        errorFile: errFile,
        typeFile: typeFile
       optionsFile: optionsFile
    };
}
```

lib/shell/shellscript.js
```diff
@@ -74,6 +74,18 @@ class ShellScriptWorker {
...
+
+            if (await fs.pathExists(preparedFiles.optionsFile)) {
+                try {
+                    const options = await fs.readJSON(preparedFiles.optionsFile);
+                    if (options.postProcess === true || options.postProcess === "true") {
+                        rendition.postProcess = true;
+                    }
+                // eslint-disable-next-line no-unused-vars
+                } catch (e) {
+                    // ignore error if options file is not properly formatted
```
pheenomenon, line 86: a console.log or console.error would help

lib/shell/shellscript.js
```diff
@@ -74,6 +74,18 @@ class ShellScriptWorker {
             } else {
                 console.log('No content type information file found');
             }
+
+            if (await fs.pathExists(preparedFiles.optionsFile)) {
+                try {
+                    const options = await fs.readJSON(preparedFiles.optionsFile);
+                    if (options.postProcess === true || options.postProcess === "true") {
+                        rendition.postProcess = true;
```
alexkli, line 82: The original idea was to make this general and set the [options of the worker](https://git.corp.adobe.com/nui/asset-compute-sdk/blob/master/lib/worker.js#L62).

The downside is that since it only happens after the callback ran, we have to modify it after the fact and it cannot impact options that are needed before (say `disableSourceDownload`).

And since `rendition.postProcess` is not even a normal `option` it doesn't really work out...

So yeah, maybe this is ok for now.

PR #49, opened 2020-08-28
NUI-813 Decide if post processing should run

alexkli:
For [NUI-813](https://jira.corp.adobe.com/browse/NUI-813).

To be merged into mob/NUI-668 main PR for post processing.

- added a new `needsImagePostProcess()` function in `image.js` which decides when the post processing should run, by inspecting the intermediated rendition and looking at the instructions
  1. if the output format (`fmt`) or format of the intermediate file is not supported, it will not post process, but succeed the rendition. Nothing we can do here & assuming the worker output the right file, say xmp or non-image stuff.
  2. if there is any custom instruction like `dpi` or soon `watermark`, post processing must kick in
  3. if a worker already handles `fmt`, `width`, `height` and `quality` (in case of jpg) and there is no other instruction, then post processing is skipped. AFAICS this is the maximum our workers support today.
  4. if there is an error inspecting the intermediate rendition, the rendition will fail with `GenericError`
- logic for 1. was in there before but in `worker.js` `shouldPostProcess()`, I just moved it to be together with other format checks
- introduced global constant lists/sets such as `SUPPORTED_OUTPUT_FMT` and `EXTRA_INSTRUCTIONS`. we will have to update them as we add more features/file format support
- the code has to inspect the intermediate image file to make most decisions. in `getImageMetadata()` I am using the "json" conversion output of imagemagick for that, which gives the entire metadata of an image in one go, which can then easily be parsed in js code. to optimize the speed I am using the `[1x1+0+0]` which works fine - and it's noticeable faster the larger the image: `convert infile[1x1+0+0] json:-` in my local runs that usually takes around 50 to 100ms, plus/minus, so seems ok, as we save time by not doing an actual `convert` with an image which can take more than a second.
- added various `test-worker` tests to cover all cases (I hope - unfortunately test coverage is missing)
  - some tests (`post-skip-*`) will check for the cases where post processing should not run. to validate that, I look at a `comment` metadata field that is set by post processing (but only ever in test mode)
  - for these the test worker itself will run an imagemagick command to simulate a worker "tool" which can be controlled via an `imagemagick` field in the `params.json`
- fixed mocha postprocessing tests which did not generate valid binaries, but I could simplify by simply comparing against the `test/files/generatedFileSmall.*` files that the mocked test uses (no need to duplicate these files as base64 strings in code)
- the change to copy `/in` in test mode was creating issues as it was sometimes randomly failing in 1 or 2 test cases. very irregularly, but often enough to be annoying. the problem is the `fse.copy("/in", directories.in)`. I could mitigate it by retrying it in case it fails, since then I never had a failed test because of this again. I can see in the logs it's still happening sometimes, but the retry always works


alexkli:
The `imagePostProcess` mocha tests are failing travis ci: https://travis-ci.com/github/adobe/asset-compute-sdk/jobs/378715215

They work fine locally. Added verbose logs to see more details.

alexkli:
Problem seems to be the same for all tests - `getImageMetadata()` returning nothing:

```
post-processing rendition0.jpg failed after 0.117: Error: Reading metadata from intermediate rendition failed: Cannot read property 'image' of undefined
1228    at getImageMetadata (/home/travis/build/adobe/asset-compute-sdk/lib/postprocessing/image.js:26:88)
1229    at process._tickCallback (internal/process/next_tick.js:68:7)
```

Maybe an older version of imagemagick on travis ci that does not support `json:-`?
https://travis-ci.com/github/adobe/asset-compute-sdk/jobs/378723650

alexkli:
Issue is fixed - was a slight different json output structure in the older imagemagick (no array at the root).

lib/postprocessing/image.js
```diff
@@ -16,10 +16,163 @@ const gm = require("./gm-promisify");
...
+    "tiff": "TIFF",
+    "gif": "GIF"
+};
+
+// list of supported input formats (intermediate rendition)
+// in imagemagick format https://imagemagick.org/script/formats.php
+const SUPPORTED_INPUT_FMT = new Set([
+    "PNG",
+    "JPEG",
+    "JPEG",
```
tmathern, line 35: Did you mean JPG? Because the set will remove the duplicatr.

lib/postprocessing/image.js
```diff
@@ -16,10 +16,163 @@ const gm = require("./gm-promisify");
...
+    "TIFF",
+    "TIFF",
```
tmathern, line 37: Did you mean TIF? Because the set will remove the duplicate.

lib/postprocessing/image.js
```diff
@@ -16,10 +16,163 @@ const gm = require("./gm-promisify");
...
+    "GIF",
+    "BMP"
+]);
+
+// minimal instructions which many workers can already handle themselves
+// const BASIC_INSTRUCTIONS = new Set([
```
tmathern, line 43: Remove it?

lib/postprocessing/image.js
```diff
@@ -16,10 +16,163 @@ const gm = require("./gm-promisify");
...
+const EXTRA_INSTRUCTIONS = new Set([
+    "crop",
+    "jpegSize",
+    "interlace",
+    "dpi",
+    "convertToDpi"
+]);
+
+function isSupportedOutputFormat(instructions) {
+    return Object.keys(SUPPORTED_OUTPUT_FMT).includes(instructions.fmt);
```
tmathern, line 60: `SUPPORTED_OUTPUT_FMT.has(nstructions.fmt);` will do the same and use the Set API.

lib/postprocessing/image.js
```diff
@@ -16,10 +16,163 @@ const gm = require("./gm-promisify");
...
+        } else {
+            return json.image;
+        }
+    } catch (e) {
+        throw new Error(`Reading metadata from intermediate rendition failed: ${e.message}`);
+    }
+}
+
+function isTargetFormat(metadata, instructions) {
+    const format = SUPPORTED_OUTPUT_FMT[instructions.fmt];
```
tmathern, line 89: `SUPPORTED_OUTPUT_FMT.has(instructions.fmt);` will do the same and use the Set API.

lib/postprocessing/image.js
```diff
@@ -16,10 +16,163 @@ const gm = require("./gm-promisify");
...
+    if (format) {
```
tmathern, line 90: `if(SUPPORTED_OUTPUT_FMT.has(instructions.fmt))` will do the same and use the Set API.

lib/postprocessing/image.js
```diff
@@ -16,10 +16,163 @@ const gm = require("./gm-promisify");
...
+}
+
+function hasExtraInstructions(instructions) {
+    return Object.keys(instructions).some(key => EXTRA_INSTRUCTIONS.has(key));
+}
+
+async function getImageMetadata(file) {
+    try {
+        // [1x1+0+0] is a trick to avoid reading all pixels into memory, improving time for large images
+        const result = await gm(`${file}[1x1+0+0]`).write("json:-");
```
jdelbick, line 74: we do this command also now in worker-flite. would be nice to store this information somewhere (maybe in a nice JSON with only the important information, maybe on the rendition) so we can re-use this inside worker-flite and won't have to call the command twice (although the command is pretty fast)

lib/postprocessing/image.js
```diff
@@ -16,10 +16,163 @@ const gm = require("./gm-promisify");
...
+        }
+    } catch (e) {
+        throw new Error(`Reading metadata from intermediate rendition failed: ${e.message}`);
+    }
+}
+
+function isTargetFormat(metadata, instructions) {
+    const format = SUPPORTED_OUTPUT_FMT[instructions.fmt];
+    if (format) {
+        return metadata.format === format;
```
tmathern, line 91: Do we need to ensure the same casing?

lib/postprocessing/image.js
jdelbick: for reference: https://git.corp.adobe.com/nui/worker-flite/blob/aa8b2a841a8b46ba1ac7d726e169a1b2173fa72d/action/worker.js#L143-L158

lib/postprocessing/image.js
```diff
@@ -16,10 +16,163 @@ const gm = require("./gm-promisify");
...
+        return dimension.height === instructions.height;
+    }
+
+    // if no width or height is set, we assume nothing to be done,
+    // as rendition should be original size which we do not know here
+    return true;
+}
+
+function hasJpegQuality(metadata, instructions) {
+    if (metadata.format === "JPEG" && instructions.quality) {
```
tmathern, line 121: Is the format normalized? Or do we need to check for both JPG and JPEG?
Also, do we need to ensure metadata.format is upper case?

lib/postprocessing/image.js
```diff
@@ -16,10 +16,163 @@ const gm = require("./gm-promisify");
...
+    // verify if the requested format is one we can output from post-processing
+    if (!isSupportedOutputFormat(instructions)) {
+        console.log(`skipping post-processing because output fmt '${instructions.fmt}' is not supported`);
+        return false;
+    }
+
+    await logImageMagickVersion();
+
+    // inspect intermediate rendition
+    const metadata = await getImageMetadata(rendition.path);
```
tmathern, line 140: Paranoid me would verify rendition existence beforehand.

lib/postprocessing/image.js
jdelbick: nevermind, both are needed. This is examining the rendition metadata...in flite we examine the source metadata

lib/worker.js
```diff
@@ -299,44 +298,36 @@ class AssetComputeWorker {
...
-        if (!rendition.postProcess) {
-            return false;
         }
+    }
 
-        // verify if the requested format is one we can output from post-processing
-        if (!["png", "jpg", "jpeg", "gif", "tif", "tiff"].includes(rendition.instructions.fmt)) {
-            console.log(`Skipping post-processing because fmt ${rendition.instructions.fmt} is not supported`);
+    async shouldPostProcess(rendition) {
+        if (!rendition.postProcess) {
```
tmathern, line 321: Didn't we want to remove that flag and leave the decision entirely to the SDK?

test-worker/test/asset-compute/worker/post-skip-jpg-quality/params.json
```diff
@@ -0,0 +1,5 @@
+{
+    "imagemagick": "-quality 10",
```
tmathern, line 2: I thought we didn't want to expose command line args?

tmathern:
@alexkli Now it's failing because aio is missing on the machine running dependent_module_tests... (I thought it was there, obviously it isn't): 
`Running tests in test/asset-compute/worker
/bin/sh: 1: aio: not found`

lib/postprocessing/image.js
```diff
@@ -16,10 +16,163 @@ const gm = require("./gm-promisify");
...
+        // [1x1+0+0] is a trick to avoid reading all pixels into memory, improving time for large images
+        const result = await gm(`${file}[1x1+0+0]`).write("json:-");
+        const json = JSON.parse(result.toString());
+
+        // newer imagemagick returns an array (for multi-page formats), older just one object (6.8.9-9 at least)
+        if (Array.isArray(json)) {
+            return json[0].image;
+        } else {
+            return json.image;
+        }
```
pheenomenon, line 82: is the assumption here that in case of multi-page format the first index will give us all the metadata we care for ?

tmathern: (In workers we only handle the first page - that's where I think the assumption comes from. It would be easy to extend with an additional index in function parameters, if needed).

lib/worker.js
jdelbick: I think it would be nice to keep a flag that ensures it goes straight to post-processing. This would save time for `worker-flite` that already does the check inside the worker to see if it should do post-processing. No, need to examine the file twice.

lib/postprocessing/image.js
alexkli: Good catch, it's an accidental duplicate.

lib/postprocessing/image.js
alexkli: Good catch, it's an accidental duplicate.

lib/postprocessing/image.js
alexkli: `SUPPORTED_OUTPUT_FMT` is an object, not a set. We need both our `fmt` strings and the imagemagick format strings, that's why it's an object.

lib/postprocessing/image.js
alexkli: `SUPPORTED_OUTPUT_FMT` is an object, not a set. We need both our `fmt` strings and the imagemagick format strings, that's why it's an object.

lib/postprocessing/image.js
alexkli: They run on different files: Here it runs on the intermediate rendition, whereas in worker-flite you inspect the source.

Regarding a reusable API: I think since flite is the only other case right now, "duplicating" that code once is fine. Having to add this API, support it and create some coupling here (what if on source files we need to slightly tweak the command?) seems like extra cost/risk that we might want to avoid for now.

Definitely worth thinking about if the need comes up in other places in the future.

lib/postprocessing/image.js
alexkli: The imagemagick file `format` values are fixed strings (always upper case), hence I have them explicitly in the `SUPPORTED_OUTPUT_FMT` as the values as taken from the im docs.

lib/postprocessing/image.js
tmathern: Right, I misread.

lib/postprocessing/image.js
alexkli: Yeah, I wanted to have these fields explicitly listed somehow. These describe what "maximum" we can expect already be handled by workers. Right now they are covered by the 3 functions `isTargetFormat()`, `isExpectedSize()` and `hasJpegQuality()`. I wrote them down in this Set first, but then couldn't find a good way to "force" use of it, so I commented it out for now.

Maybe just turn this list of basic operations into a comment next to `EXTRA_INSTRUCTIONS`?

lib/postprocessing/image.js
alexkli: The `format` returned by imagemagic are fixed and documented here: https://imagemagick.org/script/formats.php

lib/worker.js
alexkli: Correct, opt-in. And in the flite special case we always need a way to tell the post processing to not run... although if the logic is right, the decision in sdk should also figure out it should not run. But since we already made the check in flite, it's good to not repeat it and e.g. inspect the intermediate rendition unnecessarily.

test-worker/test/asset-compute/worker/post-skip-jpg-quality/params.json
alexkli: This is only implemented by the test worker to quickly simulate a worker tool "already" doing something (like fmt or resize): https://github.com/adobe/asset-compute-sdk/blob/b9b6e00f777895f70937e721df9a04bd1f6e2e54/test-worker/worker.js#L26

lib/postprocessing/image.js
alexkli: Yes. Current imagemagick will always return an array, with one entry for most images that only have one "page". But an older imagemagick (which is on Travis CI) returned just one object, no array.

This was only needed to make it work. We currently do not care about anything else than the first page, since the input formats here for the post processing all just have one page since they are plain raster images. (Although TIFFs could have multiple images/pages in them, but we don't need to support them).

alexkli:
@tmathern @pheenomenon We should install `aio` in the dependent-module-test CI scripts. I can take a look at it.

lib/postprocessing/image.js
alexkli: Done in https://github.com/adobe/asset-compute-sdk/pull/49/commits/9204ee092d4899bc8eb5731ca2c7861577cfcfc5

lib/postprocessing/image.js
alexkli: Done in https://github.com/adobe/asset-compute-sdk/pull/49/commits/8e39ca05fb83026841e3cb361664a7f1515a76b5

alexkli:
Added `aio` here: https://git.corp.adobe.com/nui/circleci/commit/a1be2324d14ae4121595d52898e034d91e5d1c6b

Waiting for a new `dependent_module_tests` run...

PR #52, opened 2020-09-02
use imagemagick to set colorspace to sRGB for common image types

jdelbick:
## Description

supporting PR in postprocessing to this: https://git.corp.adobe.com/nui/worker-flite/pull/66
(when we switch worker-flite to [use post-processing](https://git.corp.adobe.com/nui/worker-flite/pull/65), we don't want to lose this bug fix)

In order to cause the least amount of side effects, I only set the colorspace for jpg, jpeg, and png renditions

Fixes # https://jira.corp.adobe.com/browse/NUI-826
## Types of changes

## Checklist:


lib/postprocessing/image.js
```diff
@@ -227,6 +227,11 @@ async function applyOutputProperties(img, instructions) {
         img.define(`jpeg:extent=${instructions.jpegSize}`);
     }
 
+    // make sure colorspace is sRGB: bug with CMYK files producing CMYK images
+    if (instructions.fmt === "jpg" || instructions.fmt === "jpeg" || instructions.fmt === "png") {
+        img.colorspace('sRGB');
```
alexkli, line 232: As discussed, we should always set it regardless of output format (unless imagemagick doesn't like it for other file types?).

bobvanmanen: What does the function actually guarantee? JPEG, PNG, GIF output compatible with the OOTB workers? If so, it should also define that the result is always RGB.

alexkli: Comment explanation would be something like: "Most asset compute renditions are used in digital display context (web, screens), where RGB is required"

jdelbick: I changed it to set `sRGB` for all file formats. I also added a cmykJpg -> gif test to verify this works for gif formats.

jdelbick:
@alexkli yes, we can use the `convert "rendition.png[1x1+0+0]" json:` trick if we don't already examine that in the integration tests

lib/postprocessing/image.js
jdelbick: btw, any preference between `sRGB` and `RGB`?

alexkli:
@jdelbick We already do that: `getImageMetadata()` [returns the full metadata json](https://git.corp.adobe.com/nui/integration-tests/blob/master/test/integration.test.js#L990-L995), so colorspace could be easily verified.

PR #43, opened 2020-08-19
[STALE] NUI-748 - Update SDK to download the image used for watermarking

DrGirlfriend:
## Description

**The changes from this was merged into [asset-compute-sdk PR](https://github.com/adobe/asset-compute-sdk/pull/48)
Leaving this here for now - until all moving parts are complete. Pls dont close yet.**

-----------------------------------

Added watermark https and data url for rendition.watermark.watermarkContent
Added download of watermark
Added validation if watermark is PNG only by extension since that is currently how source is being checked

Needs refactoring to modularize common validation checks and workflows where the source object parameters are hard coded (probably some of the rendition checks also)

I'm also not sure about the current batch processing tests failing. I will take a look tomorrow unless someone is working on them - I pulled in mob/NUI-668 today

As for throwing a specific error for watermark downloads failing, should I just try/catch the download method and throw `RenditionFormatUnsupportedError` if download fails instead of the `GenericError` `http.js` and `datauri.js` do currently

[NUI-748 - Update SDK to download the image used for watermarking](https://jira.corp.adobe.com/browse/NUI-748)

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] I have signed the [Adobe Open Source CLA](https://opensource.adobe.com/cla.html).
- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #55, opened 2020-09-07
NUI-665 Watermarking feature (watermark changes after post processing)

pheenomenon:
## Description

This PR includes all the watermark changes(working- all tests pass) from previous PRs( #43 and #48 )

Outstanding sub-tasks:
[NUI-815](https://jira.corp.adobe.com/browse/NUI-815) Reuse source download for watermark download
[NUI-806](https://jira.corp.adobe.com/browse/NUI-806) Watermark review feedback from Alex
[NUI-752](https://jira.corp.adobe.com/browse/NUI-752) Add integration test for watermarking
[NUI-816](https://jira.corp.adobe.com/browse/NUI-816) Add unit tests for watermark post-processing (have added one test, but could add few more.)

Fixes # [NUI-665](https://jira.corp.adobe.com/browse/NUI-665)

## Types of changes

## Checklist:


lib/postprocessing/image.js
```diff
@@ -178,6 +181,35 @@ async function needsImagePostProcess(rendition) {
     return false;
 }
 
+async function processWatermark(img, renditionInstructions, outdir) {
+    console.log('Processing watermark...');
+
+    const watermarkProcessor = new WatermarkProcessor(renditionInstructions);
+    const resizedWatermarkPath = renditionInstructions.watermark.name ? path.join(outdir, 'resized-'+ renditionInstructions.watermark.name)
+        : path.join(outdir, 'resized-'+ path.basename(renditionInstructions.watermark.path));
```
tmathern, line 189:
```suggestion
        : path.join(outdir, 'resized-', path.basename(renditionInstructions.watermark.path));
```

tmathern: (I wouldn't mix `join` and concat).

lib/postprocessing/image.js
```diff
@@ -178,6 +181,35 @@ async function needsImagePostProcess(rendition) {
...
+
+    try {
+        await watermarkProcessor.fitWatermark(resizedWatermarkPath);
+    } catch(err) {
+        console.error('Error while resizing watermark to rendition dimensions', err);
+        // throw err; or should we revert to using original watermark as is i.e use renditionInstructions.watermark.path ?
```
tmathern, line 195: I'd suggest to throw:
If you have a very large watermark, you may actually end up being in a "transparent" part of the watermark and it would look like there is no watermark at all.

lib/postprocessing/watermarkProcessor.js
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    getWatermarkDimensions (path) {
+        const dimensions = getDimensions(path);
+        this.watermarkWidth = dimensions.width;
+        this.watermarkHeight = dimensions.height;
+    }
+
+    compareRenditionWidth () {
+        console.log("renditionWidth" + this.renditionWidth + "<=" +  "watermarkWidth" + this.watermarkWidth );
```
tmathern, line 39: Nitpick 1: 1 space too much at the end before the parenthesis - or one missing at the beginning.
Nitpick 2: Why a concat and not the string interpolation (`${my-var}... `)?

lib/postprocessing/watermarkProcessor.js
```diff
@@ -0,0 +1,114 @@
...
+        if( this.renditionWidth <= this.watermarkWidth ) {
+            return this.renditionWidth;
+        }
+        return 0;
+    }
+
+    compareRenditionHeight () {
+
+        console.log("HHHH", `rendition ${this.renditionHeight} <= water ${this.watermarkHeight}`);
```
tmathern, line 48: This looks like a debug statement that can be removed?

lib/postprocessing/watermarkProcessor.js
```diff
@@ -0,0 +1,114 @@
...
+
+        if (this.renditionHeight <= this.watermarkHeight) {
```
tmathern, line 50: Nitpick: Could become a ternary expression, also.

lib/postprocessing/watermarkProcessor.js
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    async fitWatermark(destinationAsset){
+
+        const resizedAsset = destinationAsset || this.watermarkPath.replace("watermark", "watermark-resized");
+        const img = new ImageOperations(this.watermarkPath);
+
+        this.getWatermarkDimensions(this.watermarkPath);
+        console.log("width start: " + this.watermarkWidth);
+        console.log("height start: " + this.watermarkHeight);
```
tmathern, line 63: I think this could be merged into one console.log.

lib/postprocessing/watermarkProcessor.js
```diff
@@ -0,0 +1,114 @@
...
+        if (watermarkWidthResize > 0) {
+            console.log("Resize W:", watermarkWidthResize);
+
+            img.resize(watermarkWidthResize);
+        }
+
+        await img.write(resizedAsset);
+
+        this.getWatermarkDimensions(resizedAsset);
+        console.log("width: " + this.watermarkWidth);
```
tmathern, line 76:
```suggestion
        console.log("width: ", this.watermarkWidth);
```

lib/postprocessing/watermarkProcessor.js
```diff
@@ -0,0 +1,114 @@
...
+        console.log("height: " + this.watermarkHeight);
```
tmathern, line 77:
```suggestion
        console.log("height: ", this.watermarkHeight);
```

lib/postprocessing/watermarkProcessor.js
```diff
@@ -0,0 +1,114 @@
...
+        const watermarkHeightResize = this.compareRenditionHeight();
+        if (watermarkHeightResize > 0) {
+            console.log("Resize H:", watermarkHeightResize);
+            const img = new ImageOperations(this.watermarkPath);
+            img.resize(null, watermarkHeightResize);
+            await img.write(resizedAsset);
+        }
+
+        //shrink
+        if (this.watermarkWidthPercent < 100) {
```
tmathern, line 88: `watermarkWidthPercent` can be a floating point number too, right? No validation prevents that, I think.

lib/postprocessing/watermarkProcessor.js
```diff
@@ -0,0 +1,114 @@
...
+        const resizedAsset = destinationAsset || this.watermarkPath.replace("watermark", "watermark-resized");
+        const img = new ImageOperations(this.watermarkPath);
+
+        this.getWatermarkDimensions(this.watermarkPath);
+        console.log("width start: " + this.watermarkWidth);
+        console.log("height start: " + this.watermarkHeight);
+
+        const watermarkWidthResize = this.compareRenditionWidth();
+
+        if (watermarkWidthResize > 0) {
```
tmathern, line 67: `watermarkWidthPercent` can be a floating point number too, right? No validation prevents that, I think.

lib/postprocessing/watermarkProcessor.js
```diff
@@ -0,0 +1,114 @@
...
+            img.resize(null, watermarkHeightResize);
+            await img.write(resizedAsset);
+        }
+
+        //shrink
+        if (this.watermarkWidthPercent < 100) {
+
+            this.getWatermarkDimensions(resizedAsset);
+            console.log("width: " + this.watermarkWidth);
+            console.log("height: " + this.watermarkHeight);
```
tmathern, line 92:
```suggestion
            console.log(`width: ${this.watermarkWidth}, height: ${this.watermarkHeight}`);
```

lib/postprocessing/watermarkProcessor.js
```diff
@@ -0,0 +1,114 @@
...
+
+            const percent = (this.watermarkWidth * .01);
+
+            const shrink = (this.watermarkWidthPercent * percent);
+
+            console.log("%", percent);
+            console.log("shrink %", shrink);
```
tmathern, line 99: Could become only 1 console.log line?

lib/postprocessing/watermarkProcessor.js
```diff
@@ -0,0 +1,114 @@
...
+
+            const img = new ImageOperations(resizedAsset);
+            img.resize(shrink);
+            await img.write(resizedAsset);
+        }
+
+        this.getWatermarkDimensions(resizedAsset);
+        console.log("new width: " + this.watermarkWidth);
+        console.log("new height: " + this.watermarkHeight);
```
tmathern, line 108: Ditto.

lib/storage.js
```diff
@@ -79,7 +99,48 @@ async function putRendition(rendition) {
...
+        if (!fs.existsSync(filePath)) {
+            throw new Error(`Invalid or missing local file ${params.watermarkContent}`);
+        }
+        return new Watermark(params, inDirectory);
+
+    } else {
+        const watermark = new Watermark(params, inDirectory);
+
+        if (path.extname(watermark.name).toLowerCase() !== ".png" && !watermark.url.startsWith("data:image/png")) {
+            throw new RenditionFormatUnsupportedError(`Invalid watermark format ${watermark.url}`);
```
tmathern, line 118: What if the failing part of the test is the extension name? This only logs the url.

lib/validate.js
```diff
@@ -81,10 +81,27 @@ function validateRendition(rendition, index, location){
...
+        validateWatermark(rendition.watermark);
+    }
 }
 
+function validateWatermark (paramsWatermark) {
+    // Only validates that content is an HTTPS or data URL
+    if (!process.env.WORKER_TEST_MODE) {
+
+        if (paramsWatermark.watermarkContent.startsWith('data:')) {
+            if (paramsWatermark.watermarkContent.startsWith('data:image/png') || !validDataUrl(paramsWatermark.watermarkContent)) {
```
tmathern, line 94: The condition is wrong. `image/png` is what we want.
Either the first part should be `!paramsWatermark.watermarkContent.startsWith('data:image/png')` or it should be `&&` instead of `||`. Not 100% sure what you want to achieve here.

lib/watermark.js
```diff
@@ -0,0 +1,56 @@
...
+    let name;
+    if (watermarkContent && validUrl.isUri(watermarkContent)) {
+        
+        if (watermarkContent.startsWith("data:image/png")) {
+            name = `${ASSET_BASENAME}.png`;
+        } else {
+            name = path.basename(url.parse(watermarkContent).pathname);
+        }
+    } else if (typeof watermarkContent === "string") {
+        // name = path.basename(params);
```
tmathern, line 46: Remove the commented line.

test-worker/test/asset-compute/worker/watermark-on-jpg/params.json
```diff
@@ -0,0 +1,11 @@
+{
+	"fmt": "jpg",
```
tmathern, line 2: Indent.

test-worker/test/asset-compute/worker/watermark-percent-on-jpg/params.json
```diff
@@ -0,0 +1,11 @@
+{
+	"fmt": "jpg",
```
tmathern, line 2: Indent.

test-worker/test/asset-compute/worker/watermark-percent-on-jpg/params.json
```diff
@@ -0,0 +1,11 @@
...
+    "watermark": {
+        "name": "watermark",
+        "placement": "Center",
+        "watermarkContent": "watermark.png",
+        "widthPercent": 25
```
tmathern, line 7:
```suggestion
        "widthPercent": 25.5
```
(To test float value handling).

tmathern: I would also add a negative test, where widthPercentage > 100, to validate error handling.

tmathern (changes requested):
I think there is one error in `validate.js` L94.

The rest are suggestions and/or nitpicks.

lib/postprocessing/image.js
jdelbick, line 189: this is a long/confusing line of code. maybe change it to an if statement so its more readable?

lib/postprocessing/image.js
jdelbick: +1 for throwing an error

lib/postprocessing/watermarkProcessor.js
jdelbick, line 39: any reason to keep this log? if not, i'd remove it

lib/postprocessing/watermarkProcessor.js
```diff
@@ -0,0 +1,114 @@
...
+    constructor(instructions) {
+
+        this.renditionWidth = instructions.width || this.getRenditionDimensions(instructions.path).width || 0;
+        this.renditionHeight = instructions.height || this.getRenditionDimensions(instructions.path).height || 0;
+        this.watermarkWidthPercent = instructions.watermark.widthPercent || 100;
+        this.watermarkPath = instructions.watermark.path;
+        this.getWatermarkDimensions(this.watermarkPath);
+    }
+
+    getRenditionDimensions (path) {
```
jdelbick, line 28: why does this have to be its own function?
I would just call `getDimensions` directly on line 20:
```js
renditionDimensions = getDimensions(instructions.path);
this.renditionWidth = renditionDimensions.width || 0;
this.renditionHeight = renditionDimensions.height || 0;
```

lib/postprocessing/watermarkProcessor.js
```diff
@@ -0,0 +1,114 @@
...
+        const img = new ImageOperations(this.watermarkPath);
+
+        this.getWatermarkDimensions(this.watermarkPath);
+        console.log("width start: " + this.watermarkWidth);
+        console.log("height start: " + this.watermarkHeight);
+
+        const watermarkWidthResize = this.compareRenditionWidth();
+
+        if (watermarkWidthResize > 0) {
+            console.log("Resize W:", watermarkWidthResize);
```
jdelbick, line 68: use full words in log messages so its less confusing:

```suggestion
            console.log(`Resized width:  ${watermarkWidthResize}`);
```

lib/postprocessing/watermarkProcessor.js
```diff
@@ -0,0 +1,114 @@
...
+
+        await img.write(resizedAsset);
+
+        this.getWatermarkDimensions(resizedAsset);
+        console.log("width: " + this.watermarkWidth);
+        console.log("height: " + this.watermarkHeight);
+
+        const watermarkHeightResize = this.compareRenditionHeight();
+        if (watermarkHeightResize > 0) {
+            console.log("Resize H:", watermarkHeightResize);
```
jdelbick, line 81: use full words in log messages so they are less confusing:
```suggestion
           console.log(`Resized height:  ${watermarkHeightResize}`);
```

lib/postprocessing/watermarkProcessor.js
```diff
@@ -0,0 +1,114 @@
...
+    async fitWatermark(destinationAsset){
+
+        const resizedAsset = destinationAsset || this.watermarkPath.replace("watermark", "watermark-resized");
+        const img = new ImageOperations(this.watermarkPath);
+
+        this.getWatermarkDimensions(this.watermarkPath);
+        console.log("width start: " + this.watermarkWidth);
+        console.log("height start: " + this.watermarkHeight);
+
+        const watermarkWidthResize = this.compareRenditionWidth();
```
jdelbick, line 65: pretty small amount of logic in both `compareRenditionWidth` and `compareRenditionHeight`... they could both be removed and use inline logic:
```suggestion
        const watermarkWidthResize = (this.renditionWidth <= this.watermarkWidth)? this.renditionWidth : 0;
```

jdelbick: same goes for watermarkHeightResize^

lib/validate.js
jdelbick: +1, also add a test case covering this

pheenomenon: yup correct `not data:image/png or inValidDataUrl` then throw error
 `!paramsWatermark.watermarkContent.startsWith('data:image/png')` seems appropriate here. will edit the tests too.

lib/postprocessing/watermarkProcessor.js
DrGirlfriend: `getRenditionDimensions()` may not even need to be called at all - It seemed cleaner and easier to read

lib/postprocessing/watermarkProcessor.js
DrGirlfriend: This is a debug line

lib/postprocessing/watermarkProcessor.js
DrGirlfriend: debug line

lib/postprocessing/watermarkProcessor.js
DrGirlfriend: it shouldn't be prevented and we have tests with floats

lib/postprocessing/watermarkProcessor.js
DrGirlfriend: Debug message

lib/postprocessing/watermarkProcessor.js
DrGirlfriend: Debug message

lib/postprocessing/watermarkProcessor.js
DrGirlfriend: debug message

lib/postprocessing/watermarkProcessor.js
DrGirlfriend: debug message

lib/postprocessing/watermarkProcessor.js
DrGirlfriend: correct

lib/postprocessing/watermarkProcessor.js
DrGirlfriend: debug message

lib/postprocessing/watermarkProcessor.js
DrGirlfriend: debug message

lib/postprocessing/watermarkProcessor.js
DrGirlfriend: yes

lib/storage.js
```diff
@@ -79,7 +99,48 @@ async function putRendition(rendition) {
...
+        }
+
+        if (disableDownload) {
+            console.log(`Skipping watermark file download for ${AssetComputeLogUtils.redactUrl(params.watermarkContent)}`);
```
pheenomenon, line 122: @DrGirlfriend I think we dont need `disableDownload` for watermark. we can remove this condition ? 
Ex: This causes worker-ffmpeg(which sets `disableSourceDownload=true` ) watermarking to fail.

README.md
```diff
@@ -130,6 +131,15 @@ Object containing the following attributes:
 | `name` | `string` | filename of the rendition to create |
 | `path` | `string` | Absolute path to store rendition locally (must put rendition here in order to be uploaded to cloud storage) |
 | `index` | `number` | number used to identify a rendition |
+| `watermark` | `object` | See [watermark](#watermark)
```
alexkli, line 134: This isn't part of `rendition` in the JS API. It would be part of `instructions`.

We do not have a section on post processing options in this README yet. Only a pointer to https://docs.adobe.com/content/help/en/asset-compute/using/api.html#rendition-instructions which we need to update instead. It would be nice to have a list here as well but then we have two places to update which easily goes out of sync.

lib/postprocessing/watermarkProcessor.js
alexkli: Ack with @jdelbick .

It might make sense if you have `getRenditionDimensions()` and `getWatermarkDimensions()` without the `path` arguments. And the paths are known by the object (they kind of are already). Then it would follow the object encapsulation. Right now the path is passed around again and again and nothing guarantees it is the same.

lib/postprocessing/watermarkProcessor.js
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    getRenditionDimensions (path) {
+        return getDimensions(path);
+    }
+
+    getWatermarkDimensions (path) {
+        const dimensions = getDimensions(path);
+        this.watermarkWidth = dimensions.width;
+        this.watermarkHeight = dimensions.height;
```
alexkli, line 35: This sounds like a getter - especially with another function `getRenditionDimensions()`, but instead it mutates the internal state. Suggest to refactor to be consistent or avoid use of the `get*()` name pattern.

lib/postprocessing/watermarkProcessor.js
alexkli: Ack, it makes it harder to follow the logic. It would be simpler to follow if it did

```
if (this.renditionWidth <= this.watermarkWidth) {
    img.resize(this.renditionWidth);
}
```

lib/postprocessing/watermarkProcessor.js
```diff
@@ -0,0 +1,97 @@
...
+
+        this.getWatermarkDimensions(this.watermarkPath);
+
+        const watermarkWidthResize = this.compareRenditionWidth();
+
+        if (watermarkWidthResize > 0) {
+            img.resize(watermarkWidthResize);
+        }
+
+        await img.write(resizedAsset);
```
alexkli, line 67: Why do we already write the `img` here and later below resize and write it again? And then potentially a 3rd time again. Just for the watermark resizing? Each write is one `imagemagick` invocation...

AFAICS we need to
1. identify the sizes of both rendition and watermark
2. then together with widthPercentage and placement options we know exactly how to size and place the watermark
3. load image and do a single resize
4. on the main rendition `gm` do the composite with that resized watermark file

This should be doable in one function and doesn't really need a separate class such as `WatermarkProcessor` here.

lib/storage.js
```diff
@@ -79,7 +99,43 @@ async function putRendition(rendition) {
     }
 }
 
+async function getWatermark(params, inDirectory) {
```
alexkli, line 102: FWIW, see also https://github.com/adobe/asset-compute-sdk/pull/57#pullrequestreview-484668073

lib/validate.js
```diff
@@ -81,10 +81,27 @@ function validateRendition(rendition, index, location){
             }
         }
     }
+    if (rendition.watermark && rendition.watermark.watermarkContent !== undefined && rendition.watermark.watermarkContent !== null) {
+        validateWatermark(rendition.watermark);
+    }
 }
 
+function validateWatermark (paramsWatermark) {
```
alexkli, line 89: Actually we don't need any validation here in worker, since that is too late and has no real benefit in production since it happens asynchronous. It all should be in api-process with immediate error feedback.

The existing validate code here can probably mostly be removed I think, we just didn't tackle that yet.

lib/worker.js
```diff
@@ -282,11 +282,18 @@ class AssetComputeWorker {
         try {
             if (await this.shouldPostProcess(rendition)) {
 
+                if (rendition.instructions.watermark && rendition.instructions.watermark.watermarkContent !== undefined && rendition.instructions.watermark.watermarkContent !== null) {
+                    rendition.instructions.watermark = await getWatermark(
+                        rendition.instructions.watermark,
+                        this.directories.in,
+                        this.options.disableSourceDownload
+                    );
+                }
```
alexkli, line 291: See https://github.com/adobe/asset-compute-sdk/pull/57#pullrequestreview-484668073 - such watermarking specific logic should be encapsulated in the post processing code only, i.e. `image.js` and co.

lib/postprocessing/imageOperations.js
```diff
@@ -0,0 +1,47 @@
...
+
+class ImageOperations {
+    constructor(infile) {
+        this.img = gm(`${infile}`);
+        this.path = infile;
+    }
+
+    resize(width, height) {
+
+        this.img.define('png:exclude-chunks=date');
```
alexkli, line 26: I don't think we need this for the intermediate watermark file.

IMO the whole class here can be replaced by using `gm` directly. The write promise is already handled by our `gm-promisify.js`.

alexkli (changes requested):

lib/postprocessing/image.js
```diff
@@ -178,6 +181,35 @@ async function needsImagePostProcess(rendition) {
...
+    // 2. resizedWatermarkPath has new watermark to be applied onto infile
+
+    const placement = renditionInstructions.watermark.placement || 'Center';
+    // composite with crop, orient not supported
+    // https://github.com/aheckmann/gm/issues/549
+    // img.gravity(placement);
+    // img.composite(resizedWatermarkPath);
+
+    img.out(resizedWatermarkPath);
+    img.gravity(placement);
```
alexkli, line 207: I am not sure we want to use imagemagick's [gravity](https://www.imagemagick.org/script/command-line-options.php#gravity) 1:1 in our `watermark.placement` instruction.

lib/postprocessing/image.js
```diff
@@ -280,6 +310,11 @@ async function applyOutputProperties(img, instructions) {
     if (process.env.SDK_POST_PROCESSING_TEST_MODE) {
         img.set("comment", "Generated by Adobe Asset Compute SDK post-processing.");
     }
+
+    // inject watermarking -compostie should come last
```
alexkli, line 314: Unless it doesn't work otherwise, this should happen above inside `applyOperations()` (line 215 which had the TODO).

`applyOutputProperties()` should be about the final output format - after the image has been rendered/resized/composed.

A clearly defined order will be more important the more features we add here.

test/postprocessing/watermarkProcessor.test.js
```diff
@@ -0,0 +1,346 @@
...
+        await fs.mkdirs('build/work', { recursive: true });
+    });
+
+    afterEach(() => {
+        delete process.env.WORKER_BASE_DIRECTORY;
+        delete process.env.ASSET_COMPUTE_NO_METADATA_IN_IMG;
+    });
+
+    describe('resizing', () => {
+        it('valid - square widthPercent at 100% ', async () => {
```
alexkli, line 37: Can we have all those tests as worker unit tests instead? Makes them easier to manage I guess.

lib/postprocessing/image.js
```diff
@@ -302,11 +337,13 @@ async function write(img, outfile) {
 
 // exported function
 async function imagePostProcess(rendition, outfile) {
+    const outdir = path.dirname(outfile);
```
alexkli, line 340: Actually a "working directory" should be passed in by the worker, who manages the directory structures here. (It can still be the same directory where the rendition (outfile) is created, but it is possible we want to change this in the future, and thus the decision does not belong in here).

Would name it `workDir` instead of `outdir` then.

lib/postprocessing/watermarkProcessor.js
alexkli: And we might be able to do this with a single `gm()` and single imagemagick process invocation:

https://stackoverflow.com/a/40940603/2709

just the first part of the answer leveraging `draw()`:

```
gm(_image)
    // WATERMARK - PARAM ORDER: [X Pos, Y Pos, width, height]
    .draw(['gravity SouthEast image Over 0,0 256,256 "/path/to/watermark.png"'])
    // RESIZE DIMENSIONS - PARAM ORDER: [width, height]
    .resize(2048, null)
    .write("/path/to/resized/output.jpg" ...)
```

test/postprocessing/watermarkProcessor.test.js
tmathern: FYI: If you try SDK tests on Windows, most tests in the worker fail: Windows doesn't resolve symlinks, it considers them being corrupt files. It's therefore a bit inconvenient if there would ever be users having to debug/modify the SDK from a Windows machine.
(Also tests that actually verify some validation without needing to generate an image could stay as unit tests here, couldn't they? @alexkli).

lib/postprocessing/image.js
tmathern: What alternative would you suggest? Do you have an idea in mind?

alexkli:
Replaced by #58

test/postprocessing/watermarkProcessor.test.js
alexkli: This hasn't changed: there are no plans to run SDK run on Windows, it's environment are only Linux containers.

tmathern: Ack - then please ignore my previous FYI (Context: I was thinking about people contributing to SDK using Windows).

PR #56, opened 2020-09-08
Storage path not handled properly

tmathern:
Storage path not handled properly when worker is in test mode (aka run-worker or test-worker).

## Description

When running a worker, storage path is not updated properly when post-processing is done. 
Therefore, creating the rendition fails with error:

```
{ [Error: ENOENT: no such file or directory, copyfile '<<source>>/out/post-rendition0.png' -> '<<target>>/out/rendition.png']
 errno: -2,
code: 'ENOENT',
syscall: 'copyfile',
path:
2020-09-08T21:36:42.797628800Z    '<<source>>/out/post-rendition0.png',
2020-09-08T21:36:42.797643800Z   dest: '<target>>/out/rendition.png' }
```

This happens because of the renaming to support postprocessing, we also need to use the proper paths when needed.
This happens only in test-mode, since we directly interact with the filesystem only and ONLY in test-mode.

To reproduce:
- Use a worker which has only one test.
- Use a rendition that needs any post-processing (e.g. resize).
- Use nui cli and do `run-worker`. The rendition should fail because of wrong path.

Supports fix for NUI-830.

## Types of changes

## Checklist:


lib/storage.js
```diff
@@ -58,21 +58,25 @@ async function putRendition(rendition) {
...
+        let sourcePath = rendition.path;
         if(rendition.name.includes('post-')) {
             // to verify post-processing renditions, since their names contain `post-` in the beginning
             // and asset-compute-cli command run-worker only looks at the file named `rendition*.*`
             rendition.name = rendition.name.substring(5);
             
             const newPath = path.join(rendition.directory, rendition.name);
             await fs.rename( rendition.path, newPath);
+            
+            sourcePath = newPath;
```
alexkli, line 70: I would avoid the term "source" here as to not confuse it with source files for a worker.

We could simply do `rendition.path = newPath` here. Would save the `sourcePath` var and all the changes below.

alexkli (changes requested):

lib/storage.js
tmathern: I found it more confusing with the 2 reassignments. 
But changed it since you think it's not worse than what is currently used.

lib/storage.js
```diff
@@ -73,6 +75,7 @@ async function putRendition(rendition) {
         if (rendition.instructions.name) {
             const newPath = path.join(rendition.directory, rendition.instructions.name);
             await fs.copyFile(rendition.path, newPath);
+            rendition.path = newPath; // assign newly created path
```
alexkli, line 78: Is this needed? It don't think any code in sdk looks at `rendition.path` after this line (since it doesn't upload in the WORKER_TEST_MODE case here, the rendition handling is done with this copyFile()).

alexkli (approved):

lib/storage.js
tmathern: Oh but it does.

alexkli: Where?

alexkli: Even if something reads it, why should it be updated here. This is like a different version of upload, just by doing copyFile instead of HTTP PUT. Anything the sdk would need from the rendition it can get from the existing rendition path (source of the copy).

tmathern: Some validation. You'll get a rendition_failed event if not updated (allthough the rendition was created).

tmathern: (^ This is empiric. It's what I observed while debugging a worker. Otherwise I wouldn't have done the change).

tmathern: Nevermind. It was my debug code that did some stuff (it was never in this PR). So I fixed it.

alexkli (approved):
Approving again.

PR #57, opened 2020-09-09
reuse source download for watermarking

jdelbick:
## Description

simplifies the code a bit. we can just re-use `Source` and `getSource` to download the watermark content.

Fixes # https://jira.corp.adobe.com/browse/NUI-815

## Types of changes

## Checklist:


lib/worker.js
```diff
@@ -283,10 +283,15 @@ class AssetComputeWorker {
...
+                    const watermarkObject = Object.assign({
+                        url: rendition.instructions.watermark.watermarkContent
+                    }, rendition.instructions.watermark);
+
+                    rendition.instructions.watermark = await getSource(
+                        watermarkObject,
                         this.directories.in,
-                        this.options.disableSourceDownload
+                        this.options.disableSourceDownload,
+                        'watermark'
```
tmathern, line 294: I find it confusing that 'watermark' here is the file's basename. Reading this without knowing the download code, I would think 'watermark' is an enum or constant value to indicate we are actually downloading a watermark.

tmathern: Can't we actually put this information in the watermark object?

test/storage.test.js
```diff
@@ -310,7 +310,7 @@ describe('storage.js', () => {
                 .get('/photo/elephant.png')
                 .reply(200, 'ok');
 
-            const watermark = await getWatermark(params, inDirectory);
+            const watermark = await getSource(params, inDirectory, true, 'watermark');
```
tmathern, line 313: I find it confusing that 'watermark' here is the file's basename. Reading this without knowing the download code, I would think 'watermark' is an enum or constant value to indicate we are actually downloading a watermark. Even more because there's just a 'true'boolean as param just before.

lib/source.js
```diff
@@ -20,17 +20,26 @@ const url = require('url');
 const SOURCE_BASENAME = 'source';
 
 class Source {
-    constructor(paramsSource, directory="") {
+    constructor(paramsSource, directory="", basename) {
```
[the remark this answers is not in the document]
alexkli: The default is always `undefined` in javascript. I think `basename=undefined` would read weird.

lib/source.js
```diff
@@ -20,17 +20,26 @@ const url = require('url');
...
+        // optional params for watermarking
+        if (paramsSource.watermarkContent) {
+            this.isWatermark = true;
+            this.watermarkContent = paramsSource.watermarkContent;
+            this.widthPercent = paramsSource.widthPercent;
```
alexkli, line 28: As mentioned in [my previous review](https://jira.corp.adobe.com/browse/NUI-806), this is watermark operation specific logic and should be fully decoupled from a generic source download code here. Otherwise we keep updating this whenever we add new operations like watermark that need file content, which makes it hard to maintain. Only the post processing logic (currently in `image.js`) should know about fields like `watermarkContent`.

Also important is that such extra files are only downloaded on demand when it actually gets to post processing and needs them. Not just automatically up front.

Instead we need to have

1. an API to download sources/files, based around the existing `Source` but generic so it can be used for other cases (IIUC basically the local source filename needs to be customizable by the consumer of the API)
   - the input is the string or object as used by `params.source` and what `watermarkContent` would also use as "type" to describe what to need
   - (nothing about `widthPercent` in there, that isn't needed for the reusable download part)
2. existing `getSource()` in `storage.js` uses it (like today)
3. `image.js` uses this API to download the watermark only when it actually needs it
   - it passes in the object, e.g. `rendition.instructions.watermarkContent`, so that the API never has to know about the names of those fields and it can be reused for different ones in the future
   - we might have to think about reuse of the same watermark between different renditions; maybe use the same filename then (but we need to ensure same url = same filename, different url = different filename)

alexkli (changes requested):

lib/source.js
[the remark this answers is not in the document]
tmathern: I'd actually suggest using the constant's value. Or `null`, for that matter.

alexkli:
Replaced by #58

PR #61, opened 2020-09-12
NUI-768 Refactor timers in SDK

alexkli:
For [NUI-768](https://jira.corp.adobe.com/browse/NUI-768).

Extended the Timer class to make the code easier to read and to support resuming measurements and making aggregations and sums easier. Also added some assertions in the unit tests for timing measurements.


lib/utils/timer.js
```diff
@@ -16,29 +16,91 @@ const process = require('process');
...
+        t._current = this._current;
+        t._total = this._current;
+        return t;
+    }
+
+    // sums multiple timer's currentDuration() and skips undefined values
+    static currentSum(...timers) {
+        let sum = 0.0;
+        timers.forEach(timer => {
+            const d = timer.currentDuration();
```
tmathern, line 76: This is hiding that the timers will be stopped (`currentDuration` calls stop). Is this the wanted behavior?

lib/utils/timer.js
```diff
@@ -16,29 +16,91 @@ const process = require('process');
...
+            }
+        });
+        return sum;
+    }
+
+    // sums multiple timer's totalDuration() and skips undefined values
+    static totalSum(...timers) {
+        let sum = 0.0;
+        timers.forEach(timer => {
+            const d = timer.totalDuration();
```
tmathern, line 88: This is hiding that the timers will be stopped (totalDuration calls stop). Is this the wanted behavior?

test/testutil.js
```diff
@@ -306,6 +308,48 @@ async function assertParamsWithMultipleRenditions(receivedMetrics) {
...
         eventType: "activation"
     }]);
+
+    assertDurationMetrics(receivedMetrics);
+}
+
+function assertDurationMetrics(receivedMetrics) {
+    receivedMetrics.filter(metric => metric.eventType === "rendition").forEach(metric => {
+        // must be positive values
+        assert.ok(metric.downloadDuration > 0.0);
```
tmathern, line 318: This is float comparision - How small are the values? Does `metric.downloadDuration > 0.0` returns true for `metric.downloadDuration === 0.001`?
Also, I maybe would rather write `0.000` in the test to indicate we expect 3 digits.

lib/utils/timer.js
alexkli: Yes. `currentDuration()`, `totalDuration()` all stop() the timer.

lib/utils/timer.js
alexkli: Yes.

test/testutil.js
alexkli: Yes it works for small values - they are very small in our unit tests and it works.

> we expect 3 digits

The 3 digit precision only applies when logged as string. Here it compares the value, and 0.0 is the same as 0.000.

tmathern: My suggestion was not technical, but if I don't look at documentation, I look at unit tests. Having 0.000 instead 0.0 would hint me that we have 3 digits précision.

(It doesn't change the code - logic, as you said. It's just a hint for people reading tests when looking for examples).

alexkli: But the most important use of these measurements is in our metrics where we have full floating point precision (whatever New Relic supports).

tmathern (approved):

tmathern:
Merged it so I have latest state for some source handling changes.

lib/worker.js
```diff
@@ -201,15 +196,17 @@ class AssetComputeWorker {
             }
 
         } catch (err) {
-            console.error(`worker() callback processing failed with error after ${callbackTimer.durationSec()} seconds: ${err.message || err}`);
+            this.timers.processingCallback.stop();
+            console.error(`worker() callback processing failed with error after ${this.timers.processingCallback} seconds: ${err.message || err}`);
```
jdelbick, line 200: a bit confused what `this.timers.processingCallback` will log...seems like its an timers class instance?

are you missing the `.toString()` at the end?

jdelbick (commented):
merged before I could review, but LGTM with one comment

jdelbick:
@tmathern you can always checkout this branch and start working off of this branch (if you expect a PR to be merged while you are making changes to the sdk)

lib/worker.js
jdelbick: does this work because console.error converts the inline variables to strings before logging? (and you override the toString() function to return a timer value)

tmathern: `console.log` calls `toString` behind the scene, and Alex overloaded `toString`: https://github.com/adobe/asset-compute-sdk/pull/61/files#diff-4c41b384c82b232501dd8416ea986ac3R67

See the code to see what is logged exactly, but it will be a duration.

tmathern:
@jdelbick I wanted all open PRs in master too :) (there were 3).

lib/worker.js
alexkli: Yes. Both `console.log()` and template strings (`${...}`) used here will use `toString()`. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/toString#Examples

lib/utils/timer.js
```diff
@@ -16,29 +16,81 @@ const process = require('process');
...
+        return this;
+    }
+
+    // stops the timer if it was running and returns the total time in seconds as float
+    stop() {
+        if (this._id) {
             try {
-                this.hrtimeElapsed = process.hrtime(this.id);
+                this._current = hrtimeToSeconds(process.hrtime(this._id));
+                this._total += this._current;
```
pheenomenon, line 41: @alexkli Is the idea here that `this._id` will be used only to track `start()` & `stop()` (a.k.a `this._current` is coupled with `this_.id`) whereas `this._total` runs independent of `this._id` ?

alexkli: `this._id` is the identifier for the current [process.hrtime()](https://nodejs.org/api/process.html#process_process_hrtime_time) run. `stop()` ends it and the next `start()` would start a new one.

This allows to resume the timer, which we use for measuring N number of worker callbacks and get the sum automatically through this class. Previously that sum was calculated in multiple places in worker.js which just added lots of noise to the code there.

PR #63, opened 2020-09-14
Source with auth

tmathern:
## Description

Add the possibility/options to set custom headers to get a source file/asset.
The custom headers should be set on `params.source.headers`. Those headers will then be used additionally to the headers defined by the sdk. If there is a name conflict, the custom set headers will win.

Fixes # [NUI-728](https://jira.corp.adobe.com/browse/NUI-728)

Selected design is to augment the `Source` class/object with a `headers` object to be used to set headers on the HTTP(s) request.

```
"source": {
    "url": "https://myurl.com/foo/bla",
    "headers": { 
        "Authorization": "auth-details-go-here"
    }
}
```

Example for Basic HTTP Auth:
```
"source": {
    "url": "https://myurl.com/foo/bla",
    "headers": { 
        "Authorization": "Basic <credentials in base64>"
    }
}
```

Example for Bearer token auth:
```
"source": {
    "url": "https://myurl.com/foo/bla",
    "headers": { 
        "Authorization": "Bearer <token>"
    }
}
```

## Technical requirements
- Needs to support HTTP Basic auth and Bearer token auth. No other authentication supported for now.
- Only downloading from authenticated sources is needed. Does not support uploading to external locations, even authenticated. 
- Does not support discovering which authentication method is used; will passthrough provided auth.

## Implementation notes
- Once the Source object has new fields, we will have to update https://github.com/adobe/node-fetch-retry in order to properly set headers on the request (probably in the options object we already use to set other things - like socket timeout handling). If the underlying HTTP library does not support it or doesn't support it properly, this will mean we may need to choose another underlying HTTP library.
- We also need to make sure no authentication information is logged, this means updating our Nui logutils with https://github.com/adobe/asset-compute-commons/pull/43 for instance. 
This is to be contributed to aio's logger (https://github.com/adobe/aio-lib-core-logging) in order to make sure that Firefly projects also do not log sensitive fields like auth.

## Other changes
- Renamed `Source` class to `Asset`.
- Removed content-type guessing for source asset download.

## Types of changes

## Checklist:

- [x] I have updated the documentation accordingly.


tmathern:
https://github.com/adobe/asset-compute-commons/pull/43

tmathern:
https://github.com/adobe/node-fetch-retry/pull/20

alexkli (changes requested):
I thought we did this in the post proc PR already, but I think we missed it: we should simply have the entire `params.source` as an object on the `Source`:

```
source.params

// in Source constructor
this.params = paramsSource;
```

(Good naming suggestions welcome)

`name` and `type` already existed, so we need to keep them for backwards compatibility. But going forward having one object containing the entire params.source makes it easy to access it in worker callbacks and we don't have to change it every time we add a new feature to params.source.

lib/storage/http.js
```diff
@@ -22,15 +22,22 @@ const DEFAULT_CONTENT_TYPE = 'application/octet-stream';
...
 async function download(source, file) {
     try {
-        const contentType = mime.contentType(source.name)  || DEFAULT_CONTENT_TYPE;
+        const contentType = mime.contentType(source.name) || DEFAULT_CONTENT_TYPE;
         console.log(`downloading source: ${AssetComputeLogUtils.redactUrl(source.url)} into ${file}`);
+
+        const headers = {
+            'content-type': contentType
+        };
+        if(source.headers){
```
jdelbick, line 31: I don't think we need the headers check here. You can just change line 28 to be:
```
const headers = Object.assign({
            'content-type': contentType
        }, source.headers)
```
as it will just ignore it if `source.headers` is undefined:
Screen Shot 2020-09-14 at 12 50 34 PM

jdelbick (commented):
I agree with Alex, we should add the entire source object to the Source Class. however, this may be better addressed in a different ticket

jdelbick:
question on the implementation note:

> Once the Source object has new fields, we will have to update https://github.com/adobe/node-fetch-retry in order to properly set headers on the request (probably in the options object we already use to set other things - like socket timeout handling). If the underlying HTTP library does not support it or doesn't support it properly, this will mean we may need to choose another underlying HTTP library.

I think this PR verified we don't need to make any changes to our library to support this: https://github.com/adobe/node-fetch-retry/pull/20...however, all of our tests use nock, would be good to run a non-mocked test locally just to verify this does actually work

tmathern:
Ok, so I updated the `Source` class (added getters for backwards compatibility).
https://github.com/adobe/node-fetch-retry/pull/20 adds tests with the headers, btw.

lib/storage/http.js
tmathern: I'm interested in the `console.log` in the `if` statement. We won't log Auth headers, so at least this indicates something custom/set through params came in. It will be useful to have that information if things fail.

lib/storage/http.js
```diff
@@ -22,15 +22,22 @@ const DEFAULT_CONTENT_TYPE = 'application/octet-stream';
...
+            console.log('using additional custom headers');
+            Object.assign(headers, source.headers);
```
pheenomenon, line 33: wonder if we should only pass through only `Authorization` header instead of everything. If only custom authorization header is needed then console logs could be specific: `using custom authorization headers`

tmathern: In https://git.corp.adobe.com/mathern/drafts/pull/1, we decided to use the whole headers object.
In our case it's to passthrough Auth headers, but I wouldn't exclude other headers further down the line.

pheenomenon (approved):

lib/source.js
```diff
@@ -15,10 +15,22 @@ const path = require('path');
 
 class Source {
     constructor(paramsSource, directory="", name="") {
+        this.sourceSettings = paramsSource;
```
jdelbick, line 18: so to access this inside a worker callback, it would be:
```
source.sourceSettings
```

IMO, the `source` prefix is repetitive. also, remember now this could be a watermark asset and not a source file so we should probably keep generic and just call it `settings`
wdyt?

jdelbick (approved):
Already approving but I think we should change the name of `sourceSettings`

lib/source.js
jdelbick: or even just `params`

jdelbick: we should probably change the name of this class to `Asset` since it is more general than source. (this can be done in a separate ticket)

tmathern: @jdelbick `sourceSettings` refers to the class name, which is `Source`...

jdelbick: @tmathern Yes obviously the source prefix was referring to the name of the class...that was not in question.

Here are my thoughts:
1. this `Source` class creates a source object that we pass to the rendition callback. So inside the rendition callback, (workers), someone would access source parameters like this `source.sourceSettings` which is repetitive. We could drop the `source` and just call it settings. that was we access it like `source.settings`
2. we create `Source` objects when we download watermarks now: https://github.com/adobe/asset-compute-sdk/blob/master/lib/storage.js#L47 so I was suggesting we change the name of the class from`Source` -> `Asset` because it is more general (this can be done in a separate ticket)

tmathern: 1. I already updated `sourceSettings` to `settings` :).
2. I would not do the name change here (aka create a ticket): because this removes backwards compatibility because of the name change. Which means updating every location using `Source` class. Also, it would also mean double-checking on core workers and custom workers that they didn't break (and their unit tests). Changing names like that is not a small feat anymore, since we GA'd.

lib/storage/http.js
```diff
@@ -22,15 +22,22 @@ const DEFAULT_CONTENT_TYPE = 'application/octet-stream';
 
 async function download(source, file) {
     try {
-        const contentType = mime.contentType(source.name)  || DEFAULT_CONTENT_TYPE;
+        const contentType = mime.contentType(source.name) || DEFAULT_CONTENT_TYPE;
         console.log(`downloading source: ${AssetComputeLogUtils.redactUrl(source.url)} into ${file}`);
+
+        const headers = {
```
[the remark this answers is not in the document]
alexkli: However, I highly **doubt we even have to set the `content-type` header**. After all we are just downloading a binary file.

We guess it based on the source extension name which is fragile anyway, and I don't think Azure Blob Store or S3 require it in any form. How are you supposed to know the content type before you retrieve it?

This might just be a relict from some old code that didn't yet know what to do :) Would be worth removing and testing (and integration testing) while we touch this code.

lib/storage/http.js
alexkli: We log params.source already for nui requests. Just logging the final headers would be fine.

(But see above, this all might be moot).

alexkli (changes requested):

lib/source.js
jdelbick: > Which means updating every location using Source class with the new name and access method (the getters are here only for backwards compatibility).

When we made breaking changes to the `Source` class in mob session, we realized it is only used in `storage.js` and `source.test.js` unit tests

changing the name of the class won't change the behavior in any of the rendition callbacks because we still refer to the class instance as `this.source`: https://github.com/adobe/asset-compute-sdk/blob/master/lib/worker.js#L169-L174

but I am fine leaving it as is, not too important as discussed over slack

lib/storage/http.js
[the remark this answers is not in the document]
jdelbick: I suggested this change earlier but there was some reason to log the that we were adding custom headers^

lib/asset.js
```diff
@@ -13,25 +13,36 @@
 'use strict';
 const path = require('path');
 
-class Source {
-    constructor(params, directory="", name="") {
-        this.settings = params;
-
+class Asset {
```
alexkli, line 16: +1  we just ned to make sure it doesn't break anything for workers. Looking at the small exposed [api](https://github.com/adobe/asset-compute-sdk#source) it should be ok, the class name is not visible to workers, but worth checking twice.

lib/asset.js
```diff
@@ -13,25 +13,36 @@
...
         this.path = path.join(directory, this.name);
     }
 
     get type() {
-        return this.settings.type;
+        if(this.params){
+            return this.params.type;
+        } else {
+            return undefined;
+        }
```
alexkli, line 28: This is a bit shorter:
```
return this.params && this.params.type;
```

I learned that you basically never have to use/return `undefined` explicitly in js.

lib/storage/http.js
```diff
@@ -15,27 +15,14 @@
...
-        console.log(`downloading source: ${AssetComputeLogUtils.redactUrl(source.url)} into ${file}`);
-
-        const headers = {
-            'content-type': contentType
-        };
-        if(source.headers){
-            console.log('using additional custom headers');
-            Object.assign(headers, source.headers);
-        }
+        console.log(`downloading source: ${AssetComputeLogUtils.redactUrl(asset.url)} into ${file}`);
```
alexkli, line 21: Rename `downloading source:` to `downloading asset:` as well.

And could log headers here as well, on a new line. Logging undefined if not set would be ok:

```suggestion
        console.log(`downloading asset ${AssetComputeLogUtils.redactUrl(asset.url)} into ${file}\nheaders:`, asset.headers);
```

alexkli (approved):
[Integration tests](https://circle.ci.adobeitc.com/gh/nui/ci-opensource/8277) look good regarding the removal of the `content-type` header for download.

@jdelbick it seems this header was added [in this commit](https://github.com/adobe/asset-compute-sdk/commit/38f2b62e1ac1e58e178f999e925a216242d4c1d0). While for upload it is definitely required (to set the right mime type for the new file in the blob store), it seems adding it for download was more of a copy/paste?

lib/asset.js
tmathern: We are not aware of custom workers already being used in production? THose are the only ones I can think which may drill down into the SDK and use this private API. If we don't have knowledge of any custom worker in prod, we can trust the IT tests to tell us if core workers are OK or not.

alexkli: Yes, there are many (even if most of them are test projects by consultants, they all are production for us).

tmathern: It's not. public API so it should not be a breaking change... But would you prefer to release as breaking? Is a strategy defined how we would handle such things?
(Note that I would expect our consultants to possibly drill down into the SDK to use our classes directly if they would need to. This is also why I was hesitating to do class renamings.).

alexkli: Not release as breaking. We must never get into a situation that would require a breaking change. Just make sure it's none. If it is, change it so that it is backwards compatible.

jdelbick: I have a memory of this affect meahana for some reason...like if the uploaded rendition was uploaded to cloud without a contenttype, it was displayed weirdly in the meahana ui.

calling download without content-type should not make a difference, AFAICS

jdelbick:
@alexkli regarding content-type header:

I have a memory of this effecting meahana for some reason...like if the uploaded rendition was uploaded to cloud storagw without a contenttype, it was displayed weirdly in the meahana ui.

calling `download` without content-type should not make a difference, AFAICS

lib/asset.js
jdelbick: if we think its too risky to change the name to Asset, maybe we wait and do this in a different PR.

I mean, worst-case scenario some customer uses the `Source` class for their intermediate assets inside their custom worker or something...seems like a long shot because they would really have to dig through the sdk code to get this, but who knows

tmathern:
> 
> I have a memory of this effecting meahana for some reason...like if the uploaded rendition was uploaded to cloud storagw without a contenttype, it was displayed weirdly in the meahana ui.
> 
Sound more like something that should be fixed in the devtool, if it reappears: Try to find out content-type to properly set it on the file/image tag?

jdelbick:
@tmathern yep, in fact, I think that is already fixed because we rely on the rendition name to set the asset type now: https://github.com/adobe/asset-compute-devtool/blob/master/client/src/components/RenditionDisplay/Rendition.jsx#L50-L52

alexkli:
@jdelbick Yes, for uploading it's required, no question about that. Here we _only_ discuss removing it on the download GET request.

alexkli:
@tmathern This isn't limited to the devtool: we explicitly want to set the content type when we upload to the blob store. This is useful for any client.

tmathern:
(Setting content-type for downloaded asset has been removed and all tests unit+IT pass).

PR #65, opened 2020-09-29
NUI-867 - Image post-processing: handle workers that cannot generate the rendition format

alexkli:
New Feature - For [NUI-867](https://jira.corp.adobe.com/browse/NUI-867). Required for webp (or any other new output format) support: #66.

### Overview (read this)

This turned out to be a slightly larger change 😉 The change is actually a key part of the post processing logic, but we were lucky that so far it worked without it. However, with `webp` as new format that none of the worker (tools) support, it is required. The problem is

1. if a rendition asks for `webp`
2. but a worker callback/tool can only generate `png`
3. we must not blindly pass `fmt=webp` to the worker callback but `fmt=png` instead
4. and post processing would then convert it the png to the final `webp`

Otherwise the worker callbacks would see `fmt=webp` in step 3 and bail out with errors.

This [diagram](https://git.corp.adobe.com/nui/nui/raw/master/images/asset-compute-sdk-post-processing.png) illustrates the overall flow, with the "Post-processing decision (1)" box (top left) being the new element added in this PR, implemented by [`prepareImagePostProcess()`](https://github.com/adobe/asset-compute-sdk/pull/65/files#diff-c8cfb029f1e204a8f6da461ffa1464b0R200).

This lead to a few follow up changes around the sdk, see change list below. If there are any questions, happy to go through in a short mob session.

Along the way I improved a few things mostly around logging, and some test structure improvement. Let me know if I should separate them out.

### Changes
- (main logic) update `fmt` in renditions instructions for worker callback to a different intermediate format such as `png` if needed
  - logic only kicks in if the `fmt` field is set AND it is an output format supported by image post processing, otherwise it will leave instructions untouched
  - custom workers are not affected unless they use the `fmt` field (which we should probably reserve...)
- (**worker api**) workers can now set `options.supportedRenditionFormats`
   - until that is provided by all workers, the code includes a _temporary_ [list of supported renditions for our current workers](https://github.com/adobe/asset-compute-sdk/pull/65/files#diff-c8cfb029f1e204a8f6da461ffa1464b0R53-R93)
   - this allows dependent module & integration tests to pass w/o changes to any worker
   - there is not one list that works for all. for example, `cameraraw` can only generate `jpg + tiff`, while `pdfrasterizer` only supports `png` output
  - otherwise defaults to `["png"]`
- (impl) new [preparePostProcess()](https://github.com/adobe/asset-compute-sdk/pull/65/files#diff-644d8ede25bb88427ccf4229d03bba1bR312) in `worker.js` calling a new [prepareImagePostProcess()](https://github.com/adobe/asset-compute-sdk/pull/65/files#diff-c8cfb029f1e204a8f6da461ffa1464b0R200) in `image.js` function making the new decision
  - `prepareImagePostProcess()` runs _before_ the callback
  - `needsImagePostProcess()` runs _after_ the callback and decides if the generated rendition needs post processing
  - `imagePostProcess()` called to do post processing if needed
  - there should be the opportunity to extend this for other future post processing logic (e.g. sensei) where we plugin other parts into the `Worker.preparePostProcess()` (and co) functions
- (tests) `test-worker/worker.js` simulates a worker only supporting some output formats and fails if incorrect format is requested
- (corollary) change Rendition class to support generating the path dynamically, in case `instructions.fmt` changed
  - better rendition file separation because of that dynamic path change
  - with post processing writing to `post/` directory instead of "post-" filename prefix
  - see example below
- (corollary) throw error if options file from shell script cannot be parsed instead of silently ignoring it, as in case of post processing this could have lead to returning an incorrect rendition (the intermediate one, w/o the required post processing)

Further improvements but unrelated:

- moved content type detection from rendition.js into separate module type.js to make it easier testable
- adjusted rendition tests accordingly and using mockRequire to simulate 'file' failures
- improved logging
   - now prefixing per rendition logs with `[rendition 0]` and the right index
   - logging asset download and WORKER_TEST_MODE file moving/copying
   - see example below
- also made one slow api test faster (5 sec to 600ms)

---

### Examples

Example of various rendition files (from log below):

* source: `in/file.gif`
* rendition generated by worker callback: `out/rendition0.png`
* rendition generated by post processing: `out/post/rendition0.gif`
* (WORKER_TEST_MODE: copied back into `out` folder: `out/rendition0.gif` (so that test-worker finds it))

Logging example:

```
2020-09-29T05:53:15.718556500Z worker worker Test: gif-converttodpi -----------------------------------------------
2020-09-29T05:53:15.718578200Z work directory        : /nodejsAction/KbNvAqfg/work/kfnjogzl
2020-09-29T05:53:15.718598400Z - source directory    : /nodejsAction/KbNvAqfg/work/kfnjogzl/in
2020-09-29T05:53:15.718618400Z - renditions directory: /nodejsAction/KbNvAqfg/work/kfnjogzl/out
2020-09-29T05:53:15.751385400Z WORKER_TEST_MODE: asset path: /nodejsAction/KbNvAqfg/work/kfnjogzl/in/file.gif
2020-09-29T05:53:15.751778900Z source downloaded in 0.000 seconds
2020-09-29T05:53:15.752455900Z [rendition 0] post-processing: adjusted fmt for worker callback: 'gif' -> 'png'
2020-09-29T05:53:15.753182400Z [rendition 0] original instructions: { convertToDpi: { xdpi: 48, ydpi: 48 }, fmt: 'gif' }
2020-09-29T05:53:15.753604000Z [rendition 0] generating rendition0.png...
2020-09-29T05:53:15.754057400Z [rendition 0] instructions for 'worker()' callback: { convertToDpi: { xdpi: 48, ydpi: 48 }, fmt: 'gif' }
2020-09-29T05:53:15.754482400Z [rendition 0] [test worker] copying source to rendition: /nodejsAction/KbNvAqfg/work/kfnjogzl/in/file.gif to /nodejsAction/KbNvAqfg/work/kfnjogzl/out/rendition0.png
2020-09-29T05:53:15.755380100Z [rendition 0] worker() callback generated rendition in 0.001 seconds: rendition0.png
2020-09-29T05:53:15.768133500Z [rendition 0] .....................................................................................................
2020-09-29T05:53:15.768571900Z [rendition 0] Version: ImageMagick 7.0.9-19 Q16 x86_64 2020-01-26 https://imagemagick.org
2020-09-29T05:53:15.768600600Z Copyright: © 1999-2020 ImageMagick Studio LLC
2020-09-29T05:53:15.768621900Z License: https://imagemagick.org/script/license.php
2020-09-29T05:53:15.768643300Z Features: Cipher DPC HDRI Modules OpenMP(3.1) 
2020-09-29T05:53:15.768664200Z Delegates (built-in): bzlib cairo djvu fftw fontconfig freetype gslib jng jp2 jpeg lcms ltdl lzma openexr pangocairo png ps raqm rsvg tiff webp wmf x xml zlib
2020-09-29T05:53:15.769027100Z [rendition 0] .....................................................................................................
2020-09-29T05:53:15.832995800Z [rendition 0] post-processing image rendition rendition0.png => post/rendition0.gif
2020-09-29T05:53:15.835456300Z [rendition 0] imagemagick command: convert /nodejsAction/KbNvAqfg/work/kfnjogzl/out/rendition0.png[0] -auto-orient -orient Undefined -colorspace sRGB -units pixelsperinch -resample 48x48 -set comment 'Generated by Adobe Asset Compute SDK post-processing.' /nodejsAction/KbNvAqfg/work/kfnjogzl/out/post/rendition0.gif
2020-09-29T05:53:16.088282600Z [rendition 0] post-processing rendition0.png finished successfully in 0.333 seconds
2020-09-29T05:53:16.089957300Z [rendition 0] WORKER_TEST_MODE: moving /nodejsAction/KbNvAqfg/work/kfnjogzl/out/post/rendition0.gif to /nodejsAction/KbNvAqfg/work/kfnjogzl/out/rendition0.gif
2020-09-29T05:53:16.093562700Z [rendition 0] event not sent: rendition_created {"rendition":{"convertToDpi":{"xdpi":48,"ydpi":48},"fmt":"gif"},"metadata":{},"type":"rendition_created","date":"2020-09-29T05:53:16.093Z","requestId":"Test: gif-converttodpi -----------------------------------------------","source":"file.gif"}
2020-09-29T05:53:16.127828700Z XXX_THE_END_OF_A_WHISK_ACTIVATION_XXX
```


tmathern:
two things that come to my mind:
- We need to update the supported formats;
- We need to update the documentation about postprocessing and custom workers, to explain what you formulated in the PR description.

lib/rendition.js
```diff
@@ -138,6 +93,19 @@ class Rendition {
...
+    get name() {
+        return Rendition.renditionFilename(this.instructions.fmt, this.index);
+    }
+
+    get path() {
+        return this._path || path.join(this.directory, this.name);
+    }
+
+    /** only for unit tests */
+    set path(path) {
```
tmathern, line 105: Since it's only for unit tests, I would rename the setter to `_path`, as the `_` is a usual indicator that a method is private in JS.

lib/rendition.js
```diff
@@ -271,12 +239,25 @@ class Rendition {
         return meta;
     }
 
+    changeInstructions(newInstructions) {
+        this._originalInstructions = this.instructions;
```
tmathern, line 243: I would do an Object.freeze (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze) on this._originalInstructions here too, so I'm sure those are not mistakenly changed later.
```suggestion
        this._originalInstructions = this.instructions;
        Object.freeze(this._originalInstructions);
```

lib/storage.js
```diff
@@ -54,8 +58,10 @@ async function getAsset(assetReference, directory, name, disableDownload) {
             // download http/https url into file
             const protocol = URL.parse(assetReference.url).protocol;
             if (protocol === DATA_PROTOCOL) {
+                console.log("creating asset from data url:", source.path);
                 await datauri.download(assetReference, source.path);
             } else {
+                console.log("downloading asset:", source.path);
```
tmathern, line 64: Aren't we supposed to redact asset paths and names to avoid leaking information?

lib/prepare.js
```diff
@@ -39,6 +40,7 @@ async function createDirectories() {
 
     directories.in   = path.resolve(directories.base, IN_DIRECTORY);
     directories.out  = path.resolve(directories.base, OUT_DIRECTORY);
+    directories.post = path.resolve(directories.out, POST_PROC_DIRECTORY);
```
tmathern, line 43: Can we rename it to `postprocessing`? 

In subsequent code I got confused for a second thinking this was HTTP(S) related to the `POST` verb...

lib/storage.js
```diff
@@ -104,28 +110,27 @@ async function getSource(paramsSource, inDirectory, disableSourceDownload) {
...
-        if(rendition.name.includes('post-')) {
-            // to verify post-processing renditions, since their names contain `post-` in the beginning
-            // and asset-compute-cli command run-worker only looks at the file named `rendition*.*`
-            rendition.name = rendition.name.substring(5);
-            
-            const newPath = path.join(rendition.directory, rendition.name);
-            await fs.rename( rendition.path, newPath);
-            
-            rendition.path = newPath;
+        if (rendition.directory === directories.post) {
```
tmathern, line 117: The name `directories.post` is confusing when used in a storage library that potentially could do HTTP(S) POST requests (see one of the previous comments). Suggesting to rename it from `post` -> `postprocessing`.

lib/worker.js
```diff
@@ -19,11 +19,12 @@ const { createDirectories, cleanupDirectories } = require('./prepare');
 const { getSource, putRendition } = require('./storage');
 const Rendition = require('./rendition');
 
-const { imagePostProcess, needsImagePostProcess } = require('./postprocessing/image');
+const { imagePostProcess, needsImagePostProcess, prepareImagePostProcess } = require('./postprocessing/image');
```
tmathern, line 22: Extreme nitpick - to have them appear in the same order as they get used:
```suggestion
const { prepareImagePostProcess, needsImagePostProcess, imagePostProcess } = require('./postprocessing/image');
```

lib/worker.js
```diff
@@ -53,6 +54,21 @@ function timeUntilTimeout() {
     return (process.env.__OW_DEADLINE - Date.now()) || DEFAULT_METRIC_TIMEOUT_MS;
 }
 
+function setConsoleLogPrefix(prefix) {
```
tmathern, line 57: Move this into `utils`? Or `commons`? It feels misplaced here. The worker should not have to worry about defining that logic - just use it.

lib/worker.js
```diff
@@ -225,7 +249,11 @@ class AssetComputeWorker {
         try {
             console.log(`generating all ${this.renditions.length} renditions...`);
             for (const rendition of this.renditions) {
-                console.log(rendition.instructionsForEvent());
+                setConsoleLogPrefix(`[rendition ${rendition.index}]`);
+                await this.preparePostProcess(rendition);
```
tmathern, line 253: try/catch around it. In case it ever crashes, to restore the proper console in `finally`.

lib/worker.js
```diff
@@ -259,23 +287,51 @@ class AssetComputeWorker {
...
         }
     }
 
+    async preparePostProcess(rendition) {
+        // post processing could ask for different instructions for the worker callback
+        const intermediateInstructions = await prepareImagePostProcess(rendition, this.options);
+        if (intermediateInstructions) {
+            console.log("original instructions:", rendition.instructionsForEvent());
+
+            rendition.changeInstructions(intermediateInstructions);
```
tmathern, line 318: I think it would be useful to log the updated instructions too here, so they're easy to compare.
```suggestion
            rendition.changeInstructions(intermediateInstructions);
            console.log("updated instructions:", rendition.instructionsForEvent());
```

lib/worker.js
```diff
@@ -259,23 +287,51 @@ class AssetComputeWorker {
...
+        }
+    }
+
+    async shouldPostProcess(rendition) {
+        if (!rendition.postProcess) {
```
tmathern, line 323: I would add a getter to see if `_forceImagePostProcess` is true and use it here. Because that means we're going to post-process anyway, and I think that one should have highest prio since we set it in the beginning when we change instructions to use an intermediary rendition.

test/postprocess.test.js
```diff
@@ -31,31 +34,37 @@ const RENDITION_JPG = fs.readFileSync(RENDITION_JPG_PATH);
...
+    before(() => {
+        const { needsImagePostProcess, prepareImagePostProcess } = require("../lib/postprocessing/image");
+        mockRequire("../lib/postprocessing/image", {
+            imagePostProcess: async function(intermediateRendition, rendition) {
+                console.log('mocked image post processing', intermediateRendition.path, rendition.path);
+                const instructions = rendition.instructions;
+                if (instructions.shouldFail) {
+                    throw new Error('mocked failure');
+                }
+                else if (instructions.fmt === 'jpg') {
```
tmathern, line 47: Nitpick: Formating; `else if` should be on previous line.

tmathern (commented):
A bunch of things to double check, see comments...

lib/postprocessing/image.js
```diff
@@ -32,14 +32,63 @@ const SUPPORTED_OUTPUT_FMT = {
...
-    "PNG",
-    "JPEG",
-    "TIFF",
-    "GIF",
-    "BMP"
-]);
+// maps imagemagick format ('PNG') https://imagemagick.org/script/formats.php to our fmt ('png')
+// Note: order is important and defines priority for intermediate format - first entry will be most preferred
+const SUPPORTED_INPUT_TYPES = {
+    "TIFF": "tif",  // most preferred format as it allows to preserve many features
```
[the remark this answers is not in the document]
alexkli: > Because if you loop over an object's keys, I think it's in alphanumerical order.

No, the order is actually preserved: https://www.stefanjudis.com/today-i-learned/property-order-is-predictable-in-javascript-objects-since-es2015/

It makes it a bit easier since I need an ordered map here (and it allows to use native features to access it). It could be switched to an array with "pair" objects in it in case the order is not guaranteed to be stable.

lib/postprocessing/image.js
```diff
@@ -32,14 +32,63 @@ const SUPPORTED_OUTPUT_FMT = {
...
+    "PNG":  "png",  // lossless over
+    "JPEG": "jpg",  // ..lossy
+    "GIF":  "gif",  // (should never get here in practice)
+    "BMP":  "bmp"
+};
+const SUPPORTED_INPUT_FORMAT_IMAGEMAGICK = Object.keys(SUPPORTED_INPUT_TYPES);
+const SUPPORTED_INPUT_FMT = Object.values(SUPPORTED_INPUT_TYPES);
+
+const DEFAULT_WORKER_SUPPORTED_FMT = [
```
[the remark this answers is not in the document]
alexkli: The preference is encoded in `SUPPORTED_INPUT_TYPES`, with the SDK deciding (e.g. prefer TIFF). Having the worker also give a preference (and then matching preferences) doesn't really solve any need we have right now I think.

lib/postprocessing/image.js
```diff
@@ -32,14 +32,63 @@ const SUPPORTED_OUTPUT_FMT = {
...
+    "png"
+];
+
+// NOTE: this is a temporary solution until workers have been updated to set options.supportedRenditionFormats
```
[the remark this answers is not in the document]
alexkli: It should rather be a jira ticket, which I will create.

lib/postprocessing/image.js
```diff
@@ -32,14 +32,63 @@ const SUPPORTED_OUTPUT_FMT = {
...
+
+    // ---------------------------------------------------------------
+    // below workers give us whatever is embedded as thumbnail format
+    // and it does not matter which fmt they are instructed to do (but the array must not be empty)
+
+    // https://git.corp.adobe.com/nui/worker-indesign/blob/6e97be7b0cd25d9eaf624d3a8200a1375abebad5/worker.js#L114
+    // per spec the embeded thumbnails in xmpGImg:format can only be JPEG (?)
+    //    https://github.com/adobe/xmp-docs/blob/master/XMPNamespaces/XMPDataTypes/Thumbnails.md
+    // https://git.corp.adobe.com/nui/core/blob/5e1dc52e950a38f93599f1095b07ae163c40dac2/lib/workers.js#L218
+    "worker-indesign": ["jpg"],
```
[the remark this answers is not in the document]
alexkli: See comment above - `worker-indesign` and `worker-dcx` don't generate anything but just extract an embedded thumbnail. So it's just like having that thumbnail as source. (Practically those thumbnails seem to be only jpeg for indesign (xmp) and png or jpeg for dcx).

But for the logic we need some value in the array here that is also in the `SUPPORTED_INPUT_TYPES` list.

lib/postprocessing/image.js
```diff
@@ -32,14 +32,63 @@ const SUPPORTED_OUTPUT_FMT = {
...
+    // https://git.corp.adobe.com/nui/core/blob/5e1dc52e950a38f93599f1095b07ae163c40dac2/lib/workers.js#L225
+    "worker-flite":     ["png", "jpg", "jpeg", "gif", "tif", "tiff"],
+
+    // https://git.corp.adobe.com/varya/imagecore_main/blob/9_9_linux/camera_raw/DAM_Raw_Converter/source/main.cpp#L203-L224
+    // https://git.corp.adobe.com/nui/core/blob/5e1dc52e950a38f93599f1095b07ae163c40dac2/lib/workers.js#L142
+    "worker-cameraraw": ["jpg", "jpeg", "tif", "tiff"],
+
+    // https://git.corp.adobe.com/nui/worker-pie/blob/925d73ad1abcb1eb7c6755eb544908250494bc11/action/worker.js#L58
+    // https://git.corp.adobe.com/nui/core/blob/5e1dc52e950a38f93599f1095b07ae163c40dac2/lib/workers.js#L180
+    "worker-pie":       ["png", "jpg", "jpeg", "gif", "tif", "tiff", "psd"],
```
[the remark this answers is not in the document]
alexkli: Not supported for output though, see the github links in the comments above.

lib/postprocessing/image.js
```diff
@@ -142,9 +191,45 @@ function hasJpegQuality(metadata, instructions) {
...
+            || WORKER_SUPPORTED_FMT[new OpenwhiskActionName().name]
+            || DEFAULT_WORKER_SUPPORTED_FMT;
+
+        // in case the worker cannot generate the final rendition format itself, we have to do it via post processing
+        if (!supportedFormats.includes(rendition.instructions.fmt)) {
+            // ...and tell the worker to generate a different intermediate format it can do
+            // find first format that is supported (worker defines priority order)
+            const newFormat = supportedFormats.find(fmt => SUPPORTED_INPUT_FMT.includes(fmt));
+            if (!newFormat) {
+                throw new GenericError(`Worker does not support generating a compatible image format. Requested: ${SUPPORTED_INPUT_FMT} Supported by worker: ${supportedFormats}`);
```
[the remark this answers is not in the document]
alexkli: Since it's an array it ends up being something like this:

```
Worker does not support generating a compatible image format. Requested: ["png", "jpg"] Supported by worker: ["tif"]
```

so I thought this has a visual delimiter already. No strong opinion.

lib/postprocessing/image.js
```diff
@@ -154,7 +239,12 @@ async function needsImagePostProcess(rendition) {
     await logImageMagickVersion();
 
     // inspect intermediate rendition
-    const metadata = rendition.intermediateMetadata = await getImageMetadata(rendition.path);
+    const metadata = rendition._intermediateMetadata = await getImageMetadata(rendition.path);
```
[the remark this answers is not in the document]
alexkli: Why not?

lib/rendition.js
alexkli: The reason for this change was to not have to change all the test code that does `rendition.path = ...`. I tried other variations (setting through constructor) but they don't work well and would have changed even more code.

lib/storage.js
alexkli: No, the _path_ does not include credentials. In the presigned URLs they are in the URL query parameters.

lib/prepare.js
alexkli: That would be too long IMO (paths are already long, making it hard to read in logs). I agree `post` is not very clear - any other suggestion that makes it more clear but is still short?

lib/storage.js
alexkli: Discussed above.

lib/worker.js
alexkli: Ack - same as done in `compute()`.

I was wondering if there would be any less intrusive way to (temporarily) change the log prefix, including restoring on errors. It just adds more boilerplate to the code right now.

lib/worker.js
alexkli: They are logged [later already](https://github.com/adobe/asset-compute-sdk/blob/b57ebd3506b5015166d9ba2e5e792dadcf2972c5/lib/worker.js#L207). I had that in here, but then it was just bloating especially if the instructions are larger (and include the `userData` from AEM or IT tests) and you see them basically twice in succession.

lib/worker.js
alexkli: Still the worker is in full control, if it says no postProcess, it should not happen at all. Think custom workers or flite.

lib/postprocessing/image.js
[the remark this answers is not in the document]
DrGirlfriend: Could we change the var name then if that is the case because looking at the var name makes me think what incoming types are supported by worker

lib/postprocessing/image.js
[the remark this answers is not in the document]
DrGirlfriend: harder to read for me at least when expecting a list of defined vars that I can skim up/down instead of up/down/right

lib/prepare.js
DrGirlfriend: `P_PROCESSING_DIRECTORY`
`POST_WORK_DIRECTORY`
`PROCESSING_DIRECTORY`
`SCRATCH_DIRECTORY`

and then make the actual directory name `postprocessing`?

lib/storage.js
DrGirlfriend: But what if the customer names the asset one of their new product names that was not released yet and some how someone got a hold of the asset name and then leaked it to the public? 

Apple was very particular to this and that is why they hosted IPS/IS internally for unreleased Apple products so they could setup the pages they would use at release time.

alexkli:
@tmathern 
1. We need to update the supported formats
    - => tracked at https://jira.corp.adobe.com/browse/NUI-873
2. We need to update the documentation about postprocessing and custom workers, to explain what you formulated in the PR description.
    - => nothing should change, as post processing is not available for custom workers at this point (no imagemagick available). we will document it once we get there, but then it will certainly have evolved.

lib/storage.js
alexkli: Logging here != leaked to the public.

We log this in many places and it is visible in metrics and logs/metrics on the AEM side to AEM devops. There is a limit to what we can hide without making it impossible to operate IMO.

lib/postprocessing/image.js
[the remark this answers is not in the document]
alexkli: The alternative would be this:

```js
const metadata = await getImageMetadata(rendition.path);
rendition._intermediateMetadata = metadata;
```

At least to me (in this particular case) it seems more hard to read since I have to see first that it's `metadata` on the second line as well...

lib/postprocessing/image.js
jdelbick, line 212: What about using `RenditionFormatUnsupported` error instead `GenericError` here?

lib/prepare.js
alexkli: I was referring to the directory name itself (currently `post`). Ok with renaming the `directories.*` field, and the constant is already specific `POST_PROC_DIRECTORY` (I think).

lib/postprocessing/image.js
```diff
@@ -142,9 +191,45 @@ function hasJpegQuality(metadata, instructions) {
...
+    // return undefined if no instruction change needed
+}
+
+/**
+ * Returns true if the (intermediate) rendition needs image post processing.
+ */
 async function needsImagePostProcess(rendition) {
     const instructions = rendition.instructions;
 
+    // NOTE: already checked in prepareImagePostProcess() above, no need to check again?
```
jdelbick, line 232: +1 this check seems repetitive now (honestly more confusing with it still here)

lib/postprocessing/image.js
```diff
@@ -346,18 +436,24 @@ async function write(img, outfile) {
     await img.write(outfile);
 }
 
-// exported function
-async function imagePostProcess(rendition, outfile, directories) {
-    const img = await read(rendition.path);
+/**
+ * Runs image post processing on the (intermediate) rendition, to create the final rendition.
+ * @param {*} intermediateRendition existing intermediate rendition created by the worker callback
```
jdelbick, line 441: nitpick: you could put the class `Rendition` in the type curly braces: `{Rendition}`

lib/prepare.js
jdelbick: `directories.postProc`?  it is not used that often, not the worst idea to use the full name
and +1 for still calling the actual directory `postprocessing`

lib/postprocessing/image.js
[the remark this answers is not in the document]
alexkli: Created [NUI-874](https://jira.corp.adobe.com/browse/NUI-874).

lib/postprocessing/image.js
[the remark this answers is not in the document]
alexkli: `FMT` refers to our `fmt` field, which is only for rendition formats. I thought this is clear, but happy to change.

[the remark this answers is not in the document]
alexkli: Renaming to `WORKER_SUPPORTED_OUTPUT_FMT`.

lib/postprocessing/image.js
```diff
@@ -142,9 +191,45 @@ function hasJpegQuality(metadata, instructions) {
...
+
+            console.log(`post-processing: adjusted fmt for worker callback: '${rendition.instructions.fmt}' -> '${newFormat}'`);
+
+            rendition._forceImagePostProcess = true;
+
+            return Object.assign({}, rendition.instructions, { fmt: newFormat});
+        }
+    }
+
+    // return undefined if no instruction change needed
```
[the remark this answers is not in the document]
alexkli: Yes, but we don't need the 3-way separation here. It's only change instructions vs. no change. This allows me to piggy back on Javascript's behavior of returning undefined by default.

jdelbick (commented):
I worry about custom workers setting `fmt` parameter in the rendition object...will this mess up processing for them?

jdelbick:
not going to lie, it was hard to review all these changes in a single PR @alexkli 

some more comments:
1. I could see the log changes in a separate PR
2. I could see the content-type changes in a separate PR (especially because you had to touch some unit tests with this change that add to the amount of material to review)
3. worker test cases...are these mostly covered in you supporting PR: https://github.com/adobe/asset-compute-sdk/pull/66?test cases like this (that would previously have failed):

> if a rendition asks for webp
but a worker callback/tool can only generate png
we must not blindly pass fmt=webp to the worker callback but fmt=png instead
and post processing would then convert it the png to the final webp

lib/prepare.js
alexkli: Changed code to

```
    directories.postprocessing = path.resolve(directories.out, POST_PROCESSING_DIRECTORY);
```

but kept the dir name as `post`.

lib/postprocessing/image.js
alexkli: I used `GenericError` because I see this as a _developer error_ (by the worker developer). If a worker opts into post processing, it must provide a compatible output format.

lib/worker.js
```diff
@@ -18,8 +18,9 @@ const process = require('process');
 const { createDirectories, cleanupDirectories } = require('./prepare');
 const { getSource, putRendition } = require('./storage');
 const Rendition = require('./rendition');
+const utils = require('./utils/utils');
```
jdelbick, line 21: kind of confusing to have a utils file in the `utils` folder

lib/prepare.js
alexkli: @jdelbick we cross posted :-)

Here is a real world example of the path:

```
2020-09-30T23:17:35.251212700Z [rendition 0] WORKER_TEST_MODE: moving /nodejsAction/08fMiXQm/work/kfq0fcje/out/post/rendition0.png to /nodejsAction/08fMiXQm/work/kfq0fcje/out/rendition0.png
```

It's already pretty long (and `kfq0fcje` will be activation id in production so even longer).

First and foremost I introduced the directory to avoid file name conflicts, i.e. to be able to have the same file in `out/file.png` as in `out/post/file.png`. This simplified a few things, also that we could just use Rendition() constructor without a `name` param but just change the `directory`.

It's not a directory that humans browse into since it's transient already. The logs show it at the beginning:

```
2020-09-30T23:17:35.082486900Z work directory        : /nodejsAction/08fMiXQm/work/kfq0fcje
2020-09-30T23:17:35.082712400Z - source directory    : /nodejsAction/08fMiXQm/work/kfq0fcje/in
2020-09-30T23:17:35.082988500Z - renditions directory: /nodejsAction/08fMiXQm/work/kfq0fcje/out
2020-09-30T23:17:35.083213500Z - post-processing dir : /nodejsAction/08fMiXQm/work/kfq0fcje/out/post
```

and I think it will be clear for us what `out/post` is since it's always the same. In this situation I feel we win more by having a bit shorter paths and thus slightly less parsing and scrolling when looking for _other_ interesting elements in the logs.

lib/postprocessing/image.js
alexkli: I think we still need it, since `prepareImagePostProcess()` never _denies_ post processing. It just makes adjustments to the instructions. This is actually the common case for all non-image renditions.

For example, let's say we have `fmt=xmp`. `prepareImagePostProcess()` will do nothing, and then when we get to `needsImagePostProcess()` here this is the place where we would decide to do nothing, since xmp is not something image post processing would work on.

Will remove the comment, which is indeed wrong.

lib/worker.js
alexkli: Better naming suggestion?

lib/postprocessing/image.js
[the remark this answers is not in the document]
tmathern: Oh, interesting! 
I still think it is a bit fragile (see https://stackoverflow.com/questions/30076219/does-es6-introduce-a-well-defined-order-of-enumeration-for-object-properties), though, since business logic depends on this being respected.

lib/postprocessing/image.js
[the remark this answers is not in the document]
tmathern: Ack.

lib/postprocessing/image.js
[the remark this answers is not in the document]
tmathern: Ack - it just looks a bit strange.

lib/postprocessing/image.js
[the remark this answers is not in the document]
tmathern: Ack. It's a question of taste, I think.

lib/worker.js
tmathern: Ack.

alexkli:
@jdelbick 

1. yeah, I was hoping it's related enough 😉  this new feature (more complex decisions) required more and better logs after all... FWIW, initially I had this together with #66 ... then I split it up already so the pure webp change is isolated.
2. the content-type changes where triggered by `Rendition.path` being a generated property now (instead of being set in the constructor) due to the rendition file/dir separation. a bit tricky to separate them out.
3. worker test cases: this is mostly covered by [this change in the test worker](https://github.com/adobe/asset-compute-sdk/pull/65/files#diff-372d7ae33e7b72869ab310eeecd2ba86R30-R34) which makes a lot of the existing test cases fail without the functional change.

lib/worker.js
tmathern: I'm not sure we can have the worker having full control. The moment we use an intermediate rendition and update instructions, if we are not doing post-processing, what guarantees we return the rendition in the requested format, and not something half-baked?
E.g. we request a webp rendition. No worker can generate a webp rendition directly yet, so we switch instructions and ask instead for a png rendition. The worker should not know about that, but the SDK does. So, to finish and return the expected rendition, aren't we forced to run post-processing to export in the proper expected format?
The way I see/understood this, the worker should be blind to the fact that we are asking it for an intermediate rendition. The worker just wouldn't have enough information to decide alone, here... Wouldn't it?

lib/worker.js
tmathern: My 2 cents:
- Put https://github.com/adobe/asset-compute-sdk/blob/NUI-867/lib/utils/utils.js#L52 in its own file, eg ùtils/console`
- THen either move the 2 remaining functions in `util/timer` or rename the file to `duration`, dealer's choice.

alexkli: They were so small that I thought a separate file is overkill .... see the comment at the top:

```
// place for small utilities. larger parts should move into their own module.
```

lib/worker.js
alexkli: `rendition.postProcess` really is a feature flag saying "I am ready for post processing as a feature". We used it to roll out the feature selectively to our ootb workers. Now that all of them support it, we could just remove it and default to true.

However, before that we have to handle 2 things:
1. we use the flag in flite as a dynamic flag (can be true or false for different activations). this is a special case that could maybe done separately / different flag (skipPostProcessing=true, etc.)
2. we must keep it turned OFF for custom workers, because the post processing logic should never run for them at this point. also we could use a different way to detect that situation reliable, but it should be one that works with existing custom workers, otherwise it would be a breaking change if we ask them to set some flag in their code and redeploy

tmathern (approved):

lib/worker.js
jdelbick: I agree it is so small that it's overkill. it's fine as is for now

lib/storage.js
```diff
@@ -104,28 +110,27 @@ async function getSource(paramsSource, inDirectory, disableSourceDownload) {
...
-            
-            const newPath = path.join(rendition.directory, rendition.name);
-            await fs.rename( rendition.path, newPath);
-            
-            rendition.path = newPath;
+        if (rendition.directory === directories.postprocessing) {
+            // move the post processing rendition to the rendition path as expected by "aio asset-compute test-worker"
+            const newRendition = new Rendition(rendition.instructions, directories.out, rendition.index);
+
+            console.log(`WORKER_TEST_MODE: moving ${rendition.path} to ${newRendition.path}`);
```
jdelbick, line 121: idea: maybe worker_test_mode also gets added to the console log prefix?

jdelbick (approved):

lib/storage.js
alexkli: IMO that would be a bit long and make it harder to read the `build/test.log`... I only added it where something special is happening in test mode, and to follow the files around (because this has become non-trivial now with all the post processing changes).

alexkli: Also the setConsoleLogPrefix() does not support nested prefixes, that would be more work to get right.

tmathern: Also, it is easy to know when we are running tests, and when we look at production logs. So there is not a lot of added value to do that here and now either.

alexkli:
More changes:

- for worker-flite: throw `RenditionFormatUnsupportedError` if worker callback skips processing (`rendition.postProcess.skippedProcessing == true`) and output format is not supported by post-processing
  - tested by new test case `skipped-unsupported-output`
- ensure a `GenericError` is thrown if no asset compute specific error is used
- only use `console.log()` so that the `[rendition N]` log prefix shows everywhere

alexkli:
@tmathern @jdelbick Merging this for the next steps with webp (i.e. #66 and testing), but not releasing yet, so we can address any feedback on the additional changes added since you all reviewed https://github.com/adobe/asset-compute-sdk/pull/65/commits/24af07b9a54d834c73f61afbffc023440d64894d.

PR #66, opened 2020-09-29
NUI-833 Support WEBP renditions

alexkli:
New Feature - For [NUI-833](https://jira.corp.adobe.com/browse/NUI-833).

- add `webp` to supported list of rendition formats in post processing logic (`SUPPORTED_OUTPUT_FMT`)
  - the rest is automatic since `imagemagick` on our base image already supports `webp`
  - default webp encoding options are ok (see jira issue), we can add optional webp encoding later
- added test cases for png to webp, one with resizing
- had to fix [validate-image.sh](https://github.com/adobe/asset-compute-sdk/pull/66/files#diff-e70a71016573795bfdbfd187d08da136) as Travis CI's imagemagick is too old and doesn't support webp (making the test validation fail). this introduced use of the [v4tech/imagemagick](https://hub.docker.com/r/v4tech/imagemagick) image, but only for tests. works well, is tiny (14.4MB) and seems popular.
- if `file` command returns `application/octet-stream`, try if `file-type` library finds a more specific type.`file` in current `worker-flite` image does not detect `webp`, but `file-type` does
- `worker-flite` special case: do not run flite at all by avoiding to adjust `fmt` for callback in the `webp` (or similar) case
- small code improvements after #65

~Before we roll this out to workers, we must first merge #65, otherwise workers will fail. Hence this PR should be merge after #65.~ Done.

Other NUI-833 PRs:
- core: https://git.corp.adobe.com/nui/core/pull/221
- IT tests: https://git.corp.adobe.com/nui/integration-tests/pull/96


alexkli:
On Travis CI, the two new test cases are failing

```
 - png-webp
      ✖  Failure: Rendition 'rendition0.webp' not as expected. Validate exit code was: 1. Check build/test-results/test-worker/test.log. 0s 285ms
 - png-webp-resize
      ✖  Failure: Rendition 'rendition0.webp' not as expected. Validate exit code was: 1. Check build/test-results/test-worker/test.log. 0s 280ms
```

They run fine locally, not quite sure what's going on. Will have to dump the test.log in Travis CI to find out more.

alexkli:
To get the logging, I added this to the `.travis.yml`:

```yaml
after_script:
  - cat test-worker/build/test-results/test-worker/test.log
```

Turns out the `webp` generation works fine according to the [logs](https://travis-ci.com/github/adobe/asset-compute-sdk/jobs/392702670), but the validation script fails, because this runs on the Travis build worker itself and relies on the imagemagick version installed there, but which is old and does not support webp:

```
Version: ImageMagick 6.8.9-9 Q16 x86_64 2019-11-12 http://www.imagemagick.org

...

--------------------------------------------------------------------------------
Running validation: '/home/travis/build/adobe/asset-compute-sdk/test-worker/test/asset-compute/worker/png-webp/validate' '/home/travis/build/adobe/asset-compute-sdk/test-worker/test/asset-compute/worker/png-webp/rendition.webp' '/home/travis/build/adobe/asset-compute-sdk/test-worker/build/test-worker/worker/2020_09_29T18_30_24_712Z/out/rendition0.webp'
...
++identify -format %wx%h /home/travis/build/adobe/asset-compute-sdk/test-worker/test/asset-compute/worker/png-webp/rendition.webp
identify: delegate failed `"dwebp" -pam "%i" -o "%o"' @ error/delegate.c/InvokeDelegate/1310.
identify: unable to open image `/tmp/magick-6800j9p5Ju4A8eVj': No such file or directory @ error/blob.c/OpenBlob/2712.
identify: unable to open file `/tmp/magick-6800j9p5Ju4A8eVj': No such file or directory @ error/constitute.c/ReadImage/540.
```

One solution could be to switch to a docker image with imagemagick for invoking `identify` here in the validate scripts, so that we have a reliable version.

alexkli:
Fixed the issue by changing [validate-image.sh](https://github.com/adobe/asset-compute-sdk/pull/66/files#diff-e70a71016573795bfdbfd187d08da136) to use imagemagick in docker.

But only in case of webp renditions, otherwise it's too slow (3x slower) and it reliable failed the tests with "socket hang up" after a while (the action container stopped, maybe because too many docker containers or mounts are being created?).

This uses the [v4tech/imagemagick](https://hub.docker.com/r/v4tech/imagemagick) image which works well, is tiny (14.4MB) and seems popular.

.travis.yml
```diff
@@ -37,8 +37,10 @@ jobs:
           branch: master
           condition:  $TRAVIS_OS_NAME == linux
         provider: script
-        skip_cleanup: true
         script: npm run semantic-release
 
+after_script:
+  - cat test-worker/build/test-results/test-worker/test.log
```
alexkli, line 43: This is useful for seeing test-worker error details.

.travis.yml
```diff
@@ -37,8 +37,10 @@ jobs:
           branch: master
           condition:  $TRAVIS_OS_NAME == linux
         provider: script
-        skip_cleanup: true
```
alexkli, line 40: true is the default and Travis constantly warns about it.

test-worker/test/asset-compute/worker/png-webp-resize/file.png
```diff
@@ -0,0 +1 @@
+../file.png
```
tmathern, line 1: Webp supports transparency. Could we add a test with transparency?
(For the matter, we should also add a test for transparency handling when doing png -> jpeg (curious to see what color jpeg will associate to the transparency) and png->gif too).

tmathern (commented):
If we look at the options (http://ftp.icm.edu.pl/pub/unix/graphics/ImageMagick/www/webp.html), is `lossless=false` the default for webp here?

We should also probably look into webp's `interlace` equivalent, progressive decoding, to offer it if interlacing is requested (see https://developers.google.com/speed/webp/faq#does_webp_support_progressive_or_interlaced_display for details). Same goes for `target-size`, which is the equivalent of the `jpegsize` option.

tmathern (approved):

alexkli:
@tmathern 
1. I believe `lossless=false` is the default. It's not well [documented](https://imagemagick.org/script/webp.php), but the underlying [webp lib's cli documentation](https://github.com/ImageMagick/webp) says `-lossless ... encode image losslessly, default=off`. Lossy by default makes sense to me for the common web page use case and as better jpeg replacement.

2. Per NUI-833 requirements we want just the default options for now. We can look at `interlace` or `targert-size` if there is demand.

   Keep it simple at this point - going from jpeg/png to default webp will be a considerable improvement. As an example, the one [test case](https://github.com/adobe/asset-compute-sdk/tree/14f905805151366a9c86401b450eeacb219f71d3/test-worker/test/asset-compute/worker/png-webp) goes from `513 KB` png to `15 KB` webp for a `1024x576` image with the same visual result (except in super fine details).

test-worker/test/asset-compute/worker/png-webp-resize/file.png
alexkli: We have some validation of transparency with the watermark tests. None for plain images here in the sdk.

I'll probably add more tests anyway and will compare with some test images. FYI, I am not merging this PR right now anyway.

.travis.yml
jdelbick: this could clog up the travis report. I would recommend only displaying the logs for the failed tests

test-worker/test/asset-compute/worker/png-webp-resize/params.json
```diff
@@ -0,0 +1,4 @@
+{
+    "width": 319,
+    "height": 319
+}
```
jdelbick, line 4: maybe I am missing something, where is the fmt param telling us to produce a webp rendition?

tmathern: (It can be implicit if the filename has the webp extension, courtesy of imagemagick)

alexkli: The expected rendition file is named `rendition.webp` and `test-worker` [uses that extension](https://github.com/adobe/aio-cli-plugin-asset-compute/blob/master/src/lib/testrunner.js#L292-L295) if it's not in the `params.json`. This makes it easier for cases without extra params, as you don't need a `params.json` file at all.

.travis.yml
alexkli: It's not a problem, see for example https://travis-ci.com/github/adobe/asset-compute-sdk/jobs/392950549 (scroll to the very bottom).

Travis is smart and collapses it by default. I had a couple of builds now (x the nodejs variants) and there was no indication of an issue.

jdelbick: sounds good.

however, it is still pretty hard to scroll through those logs (since we have so many test cases)...this is probably more general for our test worker logs but it would be nice to have separate logs for only the failing test cases

test-worker/test/asset-compute/worker/png-webp-resize/params.json
jdelbick: nice job test-worker!

DrGirlfriend (approved):

.travis.yml
alexkli: yeah, but you can search for the test case name which works pretty well IMO.

When running/debugging locally run `... test-worker <test-case-folder>` and it will just run that test. then a `cat build/test.log` will be limited to just that test.

alexkli:
Note: the IT test environment `it-20201006-adobe-asset-compute-sdk-79bc3945dc39` has been updated with webp support in `core` for testing.

alexkli:
More changes:
- if `file` command returns `application/octet-stream`, try if `file-type` library finds a more specific type.`file` in current `worker-flite` image does not detect `webp`, but `file-type` does
- `worker-flite` special case: do not run flite at all by avoiding to adjust `fmt` for callback in the `webp` (or similar) case
- small code improvements

alexkli:
There was an issue (now fixed) with `'EACCES: permission denied, unlink'` in `test-worker` runs (not production flow), only happening on Travis CI and CircleCI (native linux environments running docker), but not on Docker for Mac.

It turned out to be a problem removing files inside the `/out/post/` directory if there are any files on cleanup _after_ each test case. I spent some time debugging it and trying to fix it, but it was difficult and the usual chmod settings didn't help.

The [final fix](https://github.com/adobe/asset-compute-sdk/pull/66/files/460dbf660753cd7111e26c7b214d1ba67552cfb4..19ee07bb4295324b60da1c75b9eca1a998ccb65e) was to make sure the `post/` directory and its contents are not copied into the `/out` directory.

Possibly we can improve on that logic (in the future?) and only copy `rendition*` files to `/out` since that's all what [test-worker is looking for](https://github.com/adobe/aio-cli-plugin-asset-compute/blob/master/src/lib/testrunner.js#L459) AFAICS. Right now I noticed thinks like `options.json` and `error/` directory from shell script workers are copied too.

alexkli:
TODOs before merging this:

- [ ] determine if the logic in `prepareImagePostProcess()` for the `worker-flite` case is right and/or could be better/simpler
   
   ideally we never run the worker callback at all (and no symlinking tricks), but that requires a larger refactoring of the post processing and a new callback function provided by the worker that lets it dynamically decide if it can handle the image or not (the [existing logic whether flite can handle it](https://git.corp.adobe.com/nui/worker-flite/blob/bd45c385f50cb6b02de725d6091ca1952148e848/action/worker.js#L207-L231)), or if that works a complete declarative description of the worker what file formats it can handle and a generic check in sdk (needs to include input and output fmt, supported instructions, color profile, output)
- [ ] get new PR reviews (many new changes since the approvals)

lib/prepare.js
```diff
@@ -87,7 +87,11 @@ async function cleanupDirectories(directories) {
 
                 // ensure files can be read by host system by running chmod before copy
                 filter: src => {
+                    if (src.startsWith(directories.postprocessing)) {
+                        return false;
+                    }
                     fse.chmodSync(src, 0o766);
+                    console.log(`WORKER_TEST_MODE: copying ${src} to /out`);
```
tmathern, line 94: Is this an information about the worker's test mode or a debug leftover? Because I don't see a `copy`in the immediate neighboring lines here.

test-worker/test/asset-compute/worker/png-webp/validate
```diff
@@ -0,0 +1 @@
+../validate-image.sh
```
tmathern, line 1: I'm slightly surprised by the rendition, its quite '"pixelized".

See around the Earth atmosphere in the renditions:
image

It's because of the compression - but I don't know if this would be acceptable... But, on the other hand, we don't really have a choice here (except offering lossless webp as option).
A good thing would be to compare (losslessly compressed) tiff and (lossless) webp to see a) if there is pixelization or not and b) if the lossless webp is smaller than the tiff.

tmathern:
Note (probably for a next step):
- webp doesn't have interlacing. I'm curious to see what happens if we set this value. Could we try/add a test?
- webp's equivalent for interlacing is `progressive` which takes a value from 0 to 3. Should we support this in place of interlacing? If yes, support a new "progressive" field (preferred) or map interlacing to it (true would be progressive=3, false would be progressive=0)?

tmathern:
Already approving, but see questions/comments!

lib/asset.js
```diff
@@ -20,6 +20,10 @@ class Asset {
         this.path = path.join(directory, this.name);
     }
 
+    get extension() {
+        return this.path.split(".").pop();
```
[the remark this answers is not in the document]
alexkli: `path.extname()` includes the `.` which we do not want here.

lib/prepare.js
alexkli: The `fse.copy()` is above - this is a filter callback to decide what files in the subtree to copy or not.

The log statement is meant to stay in - to track future issues like this.

test-worker/test/asset-compute/worker/png-webp/validate
alexkli: It depends on the input file. These test files were jpgs first then turned into png. It can look better for others, and there is the `quality` option too.

In the end, we just use the standard google webp library (through imagemagick) and don't redefine the look of it.

alexkli:
@tmathern Yes, these features would be for the future. NUI-833 is limited to default webp (plus `quality`).

Imagemagick's `-interlace` doesn't seem to make a difference for webp from all I can see (not 100% sure though).

lib/postprocessing/image.js
```diff
@@ -265,7 +272,7 @@ async function needsImagePostProcess(rendition, source) {
     }
 
     // check if it can read the intermediate rendition
-    if (!isSupportedInputFormat(metadata)) {
+    if (!isSupportedInputFormat(metadata.format)) {
         console.log(`skipping post-processing because intermediate rendition format '${metadata.format}' is not supported`);
```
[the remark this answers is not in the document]
alexkli: Good question, not sure myself.

Facts:
- We get here if the output format in `fmt` would be supported (png, jpg, gif, tiff, webp) but the intermediate output format (as seen by imagemagick content type detection) is not something known.
  => I noticed there might be a bug in that the `rendition` passed into `needsImagePostProcess()` would not have the original instructions if `prepareImagePostProcess()` changed them (as in the webp case). This might affect a few things...
- It would only end up here if a worker code is faulty - generate an intermediate rendition that is not confirming to the requested `fmt`. The requested `fmt` for the intermediate rendition is now ensured by the new `prepareImagePostProcess()` (new in NUI-867)
- The logic to skip has been in [since post-processing was added](https://github.com/adobe/asset-compute-sdk/commit/ffb5248ffbd1d847ddfeafe88068f20d097f112a) and has been working fine in all tests and production so far.
- If it throws, it would fail the processing.
- If it skips (as currently), it would just pass through the result rendition of the worker as the final result.
- Changing it to throw doesn't break the tests of sdk. (Haven't checked dependent_module_tests).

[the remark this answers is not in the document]
alexkli: Yeah. I am trying out the change: https://github.com/adobe/asset-compute-sdk/pull/66/commits/cafe011e8331847874e1a3be1dc5d082636f019e

Let's see if it has any impact on workers in dependent module or IT tests.

PR #67, opened 2020-10-01
Tiff rendition to use same compression as source image, if any

tmathern:
## Description

By default, imagemagick generates lossless tiff files. Unfortunately those files are hug. This PR introduces the use of (lossless) compression for tiff files.

Fixes a bug discovered during another review.

## Types of changes

## Checklist:


lib/postprocessing/image.js
```diff
@@ -261,10 +261,12 @@ async function applyOperations(img, rendition, directories) {
     }
 }
 
-async function applyOutputProperties(img, instructions) {
-    // for reproducible png files
-    if (instructions.fmt === "png") {
+async function applyOutputProperties(img, instructions) {   
+    if (instructions.fmt === "png") { // for reproducible png files
         img.define("png:exclude-chunks=date");
+    } else if (instructions.fmt === "tif"
```
alexkli, line 267: Suggestion: to keep it easier to read and consistent, each `fmt` if condition should be in a separate `if` block like the others. There will be many over time and they will become larger. Makes it easier to read. Going by fmt makes them naturally exclusive. And have comments on their own line right above the statement:

```js
    if (instructions.fmt === "png") {
        // for reproducible png files
        img.define("png:exclude-chunks=date");
    }
    
    if (instructions.fmt === "tif" || instructions.fmt === "tiff") {
        // preserve same compression as initial file (e.g. for png -> tiff)
        img.define("tiff:preserve-compression=true");
    }
```

alexkli (commented):
Some comments below. I would like us to fully understand the tif situation and have #66 and #65 in before merging this. Generally makes sense to have a clear default on tif default compression/format, but would like to look at it more next week.

Could you symlink these to `../file.jpg`?

```
test-worker/test/asset-compute/worker/jpg-tiff-resize/file.jpg
test-worker/test/asset-compute/worker/jpg-tiff/file.jpg

```

Probably an accident by copying VS code or so, which does not preserve symlinks.

alexkli:
Failing IT tests is another occurrence of [NUI-878](https://jira.corp.adobe.com/browse/NUI-878).

test-worker/test/asset-compute/worker/jpg-tiff/params.json
```diff
@@ -0,0 +1,3 @@
+{
+    "fmt": "tif"
```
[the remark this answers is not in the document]
tmathern: I know. I just wanted 1 test where we actually have one (for covering future side-effects if we change something in the future).

tmathern:
@alexkli 
Actually it wasn't a mistake. I generated those JPEGs from the PNG example, in order to control the quality levels (this influences compression). But I can update the tests and use the file in the root. I will check the changes regarding the usefulness of the tests beforehand, though. I might even use totally different files so we actually (at least) see a notable size difference if compression is not working as expected.

lib/postprocessing/image.js
tmathern: I would actually prefer the if/else option (personal taste, I guess). THis makes the exclusion in-between expression more explicit. This would also avoid retesting all if blocks after the execution path actually went through one.

To finish, if the issue is that you expect those conditions to grow:
1. If it grows, it is going to go in its own function, something like `applyFormatOutputOptions`.
2. `if/else`usage would prevent multiple condition tests.

tmathern:
So, I changed tests to do a png -> tiff conversion. 
PNG is compressed using zlib, and with the current code, the tiff rendition is requested to use the same compression. With the change, the tif rendition has a small size (<200KB), but if ran without the change to request compression, it explodes to >3MB of size.

lib/postprocessing/image.js
```diff
@@ -261,10 +261,15 @@ async function applyOperations(img, rendition, directories) {
     }
 }
 
-async function applyOutputProperties(img, instructions) {
-    // for reproducible png files
-    if (instructions.fmt === "png") {
+async function applyOutputProperties(img, instructions) {   
+    if (instructions.fmt === "png") {
```
tmathern, line 265: Actually, I would see a better implementation being something similar to:
```
function applyFormatProperties(img, instructions){
    // for reproducible png files
    if (instructions.fmt === "png") {
        img.define("png:exclude-chunks=date");
    }

    if (instructions.fmt === "tif" || instructions.fmt === "tiff") { 
        // to keep same compression as initial file (e.g. for png -> tiff)
        img.define("tiff:preserve-compression=true");
    }

    if (instructions.fmt === "jpg" || instructions.fmt === "jpeg") {
        if(instructions.jpegSize){
            // From http://www.imagemagick.org/Usage/formats/#jpg
            // As of IM v6.5.8-2 you can specify a maximum output filesize for the JPEG image.
            img.define(`jpeg:extent=${instructions.jpegSize}`);
        }
    }
}

async function applyOutputProperties(img, instructions) {
    applyFormatProperties(img, instructions);

    // ... rest of current code

```

tmathern:
note 2: Why is this not a breaking change?
Previously, some setting ensured tiff files were compressed using a lossless algorithm. That setting was lost[1] with the update to use a custom imagemagick build. This code change is to restore the behavior[2], moving it to a visible setting in the code.

[1] The imagemagick config files before/after using the custom build are the same. Note that tests using perceptual comparison to validate the rendition will not detect size changes.
[2] Besides the exception of one worker, other workers capable of generating tiff files as renditions are not impacted by the change, as they continue to use their own code instead of imagemagick.

lib/postprocessing/image.js
alexkli: Yes, but this will just get longer and longer (e.g. we might do more inside the `png` condition), so it's nicer to separate these blocks and not force them "together" via `else if`. Also there are other blocks based on other if conditions.

alexkli (approved):
[According to this](https://github.com/ImageMagick/ImageMagick/discussions/2452#discussioncomment-54671) imagemagick changed away from the default behavior of preserving compression to lossless output by default, since that is better for some cases. They then added the `tiff:preserve-compression` option to allow the old behavior. We must have jumped across that version, hence we see the issue. Also confirms this fix is what we want to preserve Nui's standard behavior here.

tmathern:
@alexkli I'll wait for your greenlight to merge.

tmathern:
To be merged after https://github.com/adobe/asset-compute-sdk/pull/66.

DrGirlfriend (approved):

jdelbick (approved):
LGTM

PR #70, opened 2020-10-19
NUI-828 - Add SVG support to post-processing

DrGirlfriend:
NEW NOTE: [This PR](https://git.corp.adobe.com/nui/difftools/pull/13) needs to be merged and then a new docker image needs to be deployed -> `docker-project-nui-snapshot.dr.corp.adobe.com/nui/imgtools`

## Description

[NUI-828 Support SVG source files](https://jira.corp.adobe.com/browse/NUI-828)

Checks to see if source asset is SVG to allow conversion at post processing time to rendition format and saved as an intermediate file so that other post processing operations can be preformed on the converted SVG.
If rendition does not support alpha channels, a white background is forced - currently JPG's only.
Added SVG format as a valid input type.

#### Caveats: 

- ~~Fonts are not handled correctly but do render generic block characters and IM/RSVG does not throw an error failing the conversion.~~ (Handled properly for installed fonts).
- Some filters are not handled correctly (e.g. feGaussianBlur)
- External image links/URLs are not followed.

This needs to have the SVG support in our general docker image and deployed before tests will pass - [v2.4.0](https://git.corp.adobe.com/nui/openwhisk-runtime-nodejs/pull/40) or greater

**Please add `w=1` to URL parameters to bypass seeing the white space changes my editor created or use the settings widget in the Github UI**

## Types of changes

## Checklist:


lib/postprocessing/image.js
```diff
@@ -259,6 +260,10 @@ async function needsImagePostProcess(rendition, source) {
     } else if (!hasJpegQuality(metadata, instructions)) {
         console.log("running post-processing to set jpg quality");
         return true;
+
+    } else if (metadata.format === "SVG") {
+        console.log("running post-processing to convert SVG");
+        return true;
```
alexkli, line 266: This isn't necessary as the `isSupportedInputFormat()` already covers it.

lib/postprocessing/image.js
```diff
@@ -459,6 +469,35 @@ async function imagePostProcess(intermediateRendition, rendition, directories) {
...
+    }
+    image.setFormat(fmt);
+    image.rawSize(width, height); // Size of SVG before convertion
+
+    // Future - allow an instruction to define what color should be converted to transparent
+    // Future - handle fonts - currently renders generic blocks for each letter
+    // Future - inkscape - this may cover all filters currently not handled
+
+    path = path.replace(/svg*/, fmt);
+    await write(image, path);
```
alexkli, line 496: This runs a _separate_ imagemagick convert command, but we should be able to run everything in the "main" imagemagick command of post processing, by just adding the necessary extra instructions (background and rawSize). This will be faster.

alexkli (changes requested):
We should just run 1 imagemagick `convert`, not 2, which should be trivial afaics.

Otherwise LGTM, good test cases!

lib/postprocessing/image.js
DrGirlfriend: Even though some operation need to be applied in order?

lib/postprocessing/image.js
DrGirlfriend: If I read the logic correctly, checking isSupportedInputFormat() doesn't set post processing to true, checking isSupportedInputFormat() only throws and leaves the conditional block it's in, then proceeds to check if post processing is warranted in a separate conditional.


```
async function needsImagePostProcess(rendition, source) {

.......

    if (!isSupportedInputFormat(metadata.format)) {
        console.log(`Error: failing post-processing because intermediate rendition format '${metadata.format}' is not supported. The worker callback might have generated an incorrect result.`);
         throw new RenditionFormatUnsupportedError(`Format ${instructions.fmt} is not supported for ${source.extension} files`);
    }

    // check if there is an instruction that definitely needs post processing
    // fast check first, which does not requiring reading image
    if (hasExtraInstructions(instructions)) {
        console.log("running post-processing because instructions require it");
        return true;

    } else if ()....

.......

    console.log("skipping post-processing because intermediate rendition meets the target");
    return false;
}
```


```
function isSupportedInputFormat(format) {
    return SUPPORTED_INPUT_FORMAT_IMAGEMAGICK.includes(format);
}
```

alexkli: Yeah, sorry the combination of `isSupportedInputFormat()` and `isTargetFormat()` (line 252) checks should do it.

`isTargetFormat()` checks if the input (source or intermediate) is already in the output (target) format (no post processing), or not (post processing). This should always give post processing if e.g. svg -> png is being handled.

lib/postprocessing/image.js
alexkli: Imagemagick can handle pretty complex "pipelines" itself in a single command. We just have to set the commands (using `gm.xxxx()`) in the right order. We had the same with the watermarking, this was much more complex and was doable in 1 go using imagemagick's flexible `draw` command. But in this case it's trivial I think, we only have to set the background and rawSize stuff at the beginning (I think).

Probably right at the start of [`applyOperations()`](https://github.com/adobe/asset-compute-sdk/blob/master/lib/postprocessing/image.js#L319):

```js
// list of output formats not supporting alpha channels (transparency)
const NON_ALPHA_FMT = ["jpg"];

async function applyOperations(img, intermediateRendition, rendition, directories) {

    if (intermediateRendition._intermediateMetadata.format === "SVG") {
        if (NON_ALPHA_FMT.includes(fmt)) {
            // Change transparency to white
            image.background("srgb(255,255,255)");
        } else {
            // Keep transparency
            image.background("none");
        }
        // some svgs have no size (only percentage width/height), so we scale them to our target size
        image.rawSize(width, height);
    }

    // handle exif orientation
    // MUST happen before img.resize() for orientation 5 to 8 (otherwise width & height must be swapped)
    // not using img.autoOrient() because that does -strip which strips the entire exif metadata
    // but we only want to remove the exif Orientation flag which we simply unset
    img.out("-auto-orient");
    img.out("-orient");
    img.out("Undefined");

    // ...
```

I also made some more changes
- constant array `NON_ALPHA_FMT` including "jpg" and maybe more in the future
- note the use of `srgb(255,255,255)` instead of `white`. From the [colors.xml discussion](https://git.corp.adobe.com/nui/openwhisk-runtime-nodejs/pull/40#discussion_r3244843).
- some more details in the comments

alexkli: FYI, updated the above comment with `rgba(0, 0, 0, 0.0)` instead of `none`. From [here](https://git.corp.adobe.com/nui/openwhisk-runtime-nodejs/pull/40#issuecomment-2755002).

alexkli: > bit depth going from 8 to 16 ... The files are 2x as large

Would be good to understand why that is happening. I am sure this can be tweaked. I think we should avoid 2x as large.

Is this happening only for SVG input? Or does this happen for all files now? (afaics it shouldn't if we only do those extra settings if input == SVG).

> The bad news is in the scenario of JPG, the background is not forced to white (has nothing to do with the background value being named or rgba values). So I'm guessing it's part of the other ticket you mentioned last week.

What is the background result if not white? In my local testing it worked with white (using `convert -background 'rgb(255,255,255)' input.svg output.jpg`, but that was using the imagemagick on my Mac (with rsvg). So it seems possible, maybe we just have to tweak something in our imagemagick installation on the docker image from https://git.corp.adobe.com/nui/openwhisk-runtime-nodejs/pull/40.

What ticket are you referring to? (Sorry my memory doesn't find a match, but that could be me :-)

DrGirlfriend: The resulting image is all black (where the SVG image is only one color of black with transparent background)
Same result as this par of one of the SVG threads https://cq-dev.slack.com/archives/GAMJLHSBH/p1603387248125700?thread_ts=1603384448.078800&cid=GAMJLHSBH

For the ticket, I just realized I didn't create it yet.... but is related to https://cq-dev.slack.com/archives/GAMJLHSBH/p1603387788138800?thread_ts=1603384448.078800&cid=GAMJLHSBH

It looks like none of the renditions currently in the worker tests are 16 bit nor go from 8 to 16 bit. I will look some more to see where maybe the 16 bit get's introduced. Do we want to have a future feature for bit designation in the instructions object?

lib/postprocessing/image.js
```diff
@@ -459,6 +469,35 @@ async function imagePostProcess(intermediateRendition, rendition, directories) {
     await write(img, rendition.path);
 }
 
+async function convertSvg(path, instructions) {
+
+    // IM automatically keeps aspect ratio
+    const width = instructions.width || 400;
+    const height = instructions.height || 400;
```
jdelbick, line 476: one question, why is the default 400?

lib/postprocessing/image.js
alexkli: > The resulting image is all black

Would you mind sharing the SVG source image and the actual `convert` command run (if done in sdk/through test-worker, the imagemagick convert command is in the logs)? Maybe best in the jira issue or slack (not sure if the image can be shared publicly here).

> Do we want to have a future feature for bit designation in the instructions object?

Yes, possibly. But here we should keep it simple.

DrGirlfriend: So far it seems for the bit depth change:

- Writing the intermediate rendition asset does NOT convert the B&W 8 bit to 16 bit whether or not the `convertSVG()` is called in `applyOperations()` or `imagePostProcess()`
- Not writing the intermediate rendition asset DOES convert to B&W 16 bit whether or not the `convertSVG()` is called in `applyOperations()` or `imagePostProcess()`
- Defining background using `rgbs`/`rgb` vs color name does not make a difference
- This issue is not seen no all B&W SVGs

Here are the Imagemagick commands recorded in the test log maybe some sees something?

New code
```
convert -background 'rgba(0, 0, 0, 0.0)' /nodejsAction/gR2hVJjl/work/kgty20sk/out/rendition0.png[0] -size 1000x400 -auto-orient -orient Undefined -resize 1000> -define png:exclude-chunks=date -colorspace sRGB /nodejsAction/gR2hVJjl/work/kgty20sk/out/post/rendition0.png
```


Old code
```
convert -background 'rgba(0, 0, 0, 0.0)' /nodejsAction/s0BFLo15/work/kgty91pt/out/rendition0.png[0] -size 1000x400 png:-
convert /nodejsAction/s0BFLo15/work/kgty91pt/out/rendition0.png[0] -auto-orient -orient Undefined -resize 1000> -define png:exclude-chunks=date -colorspace sRGB /nodejsAction/s0BFLo15/work/kgty91pt/out/post/rendition0.png
```

alexkli: If I run this locally, I get the same results. The intermediate and the results are always 16 bit PNG, regardless of 1 or 2 step process.

If I have an 8-bit PNG as source (instead of SVG), I get an 8-bit output.

I believe this is simply due to the extra alpha channel we need to preserve the transparency.

Here is what I ran, using `lg-information.svg` from the CQ jira issue:

1. SVG one step
```
> convert -background 'rgba(0, 0, 0, 0.0)' lg-information.svg -size 1000x400 -auto-orient -orient Undefined -resize '1000>' -define png:exclude-chunks=date -colorspace sRGB out.png

> identify out.png 
out.png PNG 216x216 216x216+0+0 16-bit sRGB 13872B 0.000u 0:00.000
```

2. SVG two steps
```
> convert -background 'rgba(0, 0, 0, 0.0)' lg-information.svg -size 1000x400 intermediate.png

> identify intermediate.png 
intermediate.png PNG 216x216 216x216+0+0 16-bit sRGB 13970B 0.010u 0:00.000

> convert intermediate.png -auto-orient -orient Undefined -resize '1000>' -define png:exclude-chunks=date -colorspace sRGB out2.png

> identify out2.png 
out2.png PNG 216x216 216x216+0+0 16-bit sRGB 13872B 0.000u 0:00.000
```

3. PNG 8-bit, one step
```
> identify image.png 
image.png PNG 490x348 490x348+0+0 8-bit sRGB 196756B 0.000u 0:00.000

> convert image.png -auto-orient -orient Undefined -resize '1000>' -define png:exclude-chunks=date -colorspace sRGB out3.png

> identify out3.png 
out3.png PNG 490x348 490x348+0+0 8-bit sRGB 157082B 0.010u 0:00.000
```

alexkli: I did some more testing and the 8-bit instead of 16-bit output happens when _all_ of these are true:
1. if the SVG is pure black & white 
2. `rsvg` is used
3. no `-resize` command is present (!)
4. `-size` comes after the input file in the argument list

Regarding 2: Turns out the imagemagick on my Mac does not use `rsvg`. It's visible using `convert -list configure` (which has `--with-rsvg=no` on my mac), and `convert -list delegate` is misleading).

It seems rsvg is smart and avoids use of alpha channel or extra color bits if not needed due to the SVG being limited to black & white.

Regarding 3: I started with the full command and removed all options to identify that `-resize` makes the difference compared to the 2 step process. I think we could theoretically just have `-size` (in case of a SVG), but we would have to handle the aspect ratio logic ourselves and read the svg size first, which would not always work.

Regarding 4: This actually is not quite optimal as it means it renders the SVG onto a larger canvas first (I think?). At least if you have just `-size` and no `-resize`, the output image will have a different size depending on whether `-size` comes before the input file argument or after. To add, I am not sure if it's controllable with `gm()` which API-wise always starts with the input file.

**tldr;**

Having 16-bit output PNGs from SVG is not an issue. As long as existing conversions from PNG/JPEG etc. aren't affected, which they should not.

But we should check the `-size` position and see if we can get it right - assuming right = in front.

DrGirlfriend: Here is what I have done locally
2. I updated the docker image to install RSVG and it's dependents after new image layer is defined (checked into [PR#40](https://git.corp.adobe.com/nui/openwhisk-runtime-nodejs/pull/40)
3. `-resize` is applied outside the scope of SVG with - should I force remove if SVG? or leave as is and live with it as is
```
    if (instructions.width || instructions.height) {
        // set ">" to prevent upscaling
        // docs: https://www.imagemagick.org/script/command-line-processing.php#geometry
        img.resize(instructions.width || null, instructions.height || null, ">");
    }
```
4. Using `gm.in()` to define `-size` forces it to before the source is defined
```
convert -background 'rgba(0, 0, 0, 0.0)' -size 1000x400 /nodejsAction/7p3FcXVQ/work/kgv851kp/out/rendition0.png[0] -colorspace sRGB -auto-orient -orient Undefined -resize 1000> -define png:exclude-chunks=date -colorspace sRGB /nodejsAction/7p3FcXVQ/work/kgv851kp/out/post/rendition0.png
```

lib/postprocessing/image.js
DrGirlfriend: This was on another thread about what does AEM default to which was 400x400
Super long thread but it starts around here https://cq-dev.slack.com/archives/GAMJLHSBH/p1603232882448300?thread_ts=1603227975.430400&cid=GAMJLHSBH

lib/postprocessing/image.js
```diff
@@ -459,6 +469,35 @@ async function imagePostProcess(intermediateRendition, rendition, directories) {
...
+    const fmt = instructions.fmt || "png";
+
+    // Creates intermediate image file since IM operation order matters
+    // Note RSVG is the converter here currently 2020/10/22 causing differences between normal Imagemagic use
+    const image = await read(path);
+
+    if (fmt === "jpg") {
```
tmathern, line 483: I know we normalize `jpeg` to `jpg`, but... Better be safe.
```suggestion
    if (fmt === "jpg" || fmt === "jpeg") {
```

test-worker/test/asset-compute/worker/svg-alpha-gif/params.json
```diff
@@ -0,0 +1,4 @@
+{
+        "width" :500,
```
tmathern, line 2:
```suggestion
    "width" :500,
```
(indentation)

test-worker/test/asset-compute/worker/svg-alpha-gif/params.json
```diff
@@ -0,0 +1,4 @@
...
+        "fmt": "gif"
```
tmathern, line 3:
```suggestion
    "fmt": "gif"
```
(indentation)

test-worker/test/asset-compute/worker/svg-alpha-jpg/params.json
```diff
@@ -0,0 +1,4 @@
+{
+        "width": 500,
```
tmathern, line 2:
```suggestion
    "width": 500,
```

test-worker/test/asset-compute/worker/svg-alpha-jpg/params.json
```diff
@@ -0,0 +1,4 @@
...
+        "fmt": "jpg"
```
tmathern, line 3:
```suggestion
    "fmt": "jpg"
```

test-worker/test/asset-compute/worker/svg-png-missing-width/params.json
```diff
@@ -0,0 +1,3 @@
+{
+        "height":100
```
tmathern, line 2:
```suggestion
    "height":100
```

test-worker/test/asset-compute/worker/svg-png-missing-height/params.json
```diff
@@ -0,0 +1,3 @@
+{
+        "width":100
```
tmathern, line 2:
```suggestion
    "width":100
```

test-worker/test/asset-compute/worker/svg-aspect-ratio-width/params.json
```diff
@@ -0,0 +1,4 @@
+{
+        "width":100,
```
tmathern, line 2:
```suggestion
    "width":100,
```

test-worker/test/asset-compute/worker/svg-aspect-ratio-width/params.json
```diff
@@ -0,0 +1,4 @@
...
+        "height":500
```
tmathern, line 3:
```suggestion
    "height":500
```

test-worker/test/asset-compute/worker/svg-aspect-ratio-height/params.json
```diff
@@ -0,0 +1,4 @@
+{
+        "width":500,
+        "height":100
```
tmathern, line 3:
```suggestion
    "height":100
```

test-worker/test/asset-compute/worker/svg-aspect-ratio-height/params.json
```diff
@@ -0,0 +1,4 @@
+{
+        "width":500,
```
tmathern, line 2:
```suggestion
    "width":500,
```

test-worker/test/asset-compute/worker/svg-alpha-tiff/params.json
```diff
@@ -0,0 +1,4 @@
+{
+        "width": 500,
+        "fmt": "tiff"
```
tmathern, line 3:
```suggestion
    "fmt": "tiff"
```

test-worker/test/asset-compute/worker/svg-alpha-tiff/params.json
```diff
@@ -0,0 +1,4 @@
+{
+        "width": 500,
```
tmathern, line 2:
```suggestion
    "width": 500,
```

test-worker/test/asset-compute/worker/svg-alpha-png/params.json
```diff
@@ -0,0 +1,4 @@
+{
+        "width": 500,
+        "fmt": "png"
```
tmathern, line 3:
```suggestion
    "fmt": "png"
```

test-worker/test/asset-compute/worker/svg-alpha-png/params.json
```diff
@@ -0,0 +1,4 @@
+{
+        "width": 500,
```
tmathern, line 2:
```suggestion
    "width": 500,
```

lib/postprocessing/image.js
alexkli: 3. not sure, we should test it well. I would say it's not bad if we leave it in and get a few more 16-bit PNGs for some SVGs, hard to imagine someone would care.

4. cool 👍

lib/postprocessing/image.js
```diff
@@ -318,6 +324,12 @@ async function applyWatemark(img, intermediateRendition, rendition, directories)
 
 async function applyOperations(img, intermediateRendition, rendition, directories) {
     const instructions = rendition.instructions;
+
+    /*
+    ImageMagick supports color profiles, however, for images without a profile or a declaration of colorspace, ImageMagick assumes non-linear sRGB. Most image processing algorithms assume a linear colorspace, therefore it might be prudent to convert to linear color or remove the gamma function before certain image processing algorithms are applied.
+    */
+    // img.colorspace('RGB');
```
tmathern, line 331: @DrGirlfriend We'll need to uncomment that to use it - so we have the proper colors eventually.
But let's do this at the end, because the tests will need to be updated accordingly (since it influences the algorithms - especially resizing it seems).

lib/postprocessing/image.js
```diff
@@ -450,7 +462,13 @@ async function write(img, outfile) {
  * @param {Object} directories working directories of the current activation
  */
 async function imagePostProcess(intermediateRendition, rendition, directories) {
+
+    if (intermediateRendition._intermediateMetadata.format === "SVG") {
+        intermediateRendition.path = await convertSvg(intermediateRendition.path, intermediateRendition.instructions);
+    }
+
     const img = await read(intermediateRendition.path);
+    //img.setFormat(rendition.instructions.fmt);
```
tmathern, line 471: @DrGirlfriend I'd recommend uncommenting that too, but also at the end, when we have svg working.
(This helps with CMYK colorspace handling, by setting defaults if the format that was set can't handle CMYK).

lib/postprocessing/image.js
```diff
@@ -459,6 +479,29 @@ async function imagePostProcess(intermediateRendition, rendition, directories) {
...
+
+    // Creates intermediate image file since IM operation order matters
+    const image = await read(path);
+
+    if (fmt === "jpg" || fmt === "jpeg" ) {
+        image.background("#FFFFFF"); // Change tranparency to white 
+    } else {
+        image.background("none"); // Keep alpha background
+    }
+    image.flatten(); // forces the image on its background
```
tmathern, line 497: No intermediary write needed anymore, thanks to `flatten`!

lib/postprocessing/image.js
tmathern: (done)

lib/postprocessing/image.js
tmathern: Done in https://github.com/adobe/asset-compute-sdk/pull/77.

lib/postprocessing/image.js
```diff
@@ -39,16 +39,23 @@ const SUPPORTED_OUTPUT_FMT = {
...
+    "BMP": "bmp",
+    "SVG": "svg"
 };
 const SUPPORTED_INPUT_FORMAT_IMAGEMAGICK = Object.keys(SUPPORTED_INPUT_TYPES);
 const SUPPORTED_INPUT_FMT = Object.values(SUPPORTED_INPUT_TYPES);
+
 const DEFAULT_WORKER_SUPPORTED_OUTPUT_FMT = [
     "png"
 ];
+// list of output formats not supporting alpha channels (transparency)
```
tmathern, line 54: FYI - No need to update yet:

In another PR, I named it `FORMAT_SUPPORTS_TRANSPARENCY` (because when people wrote about TIFF, they paid attention to not name it alpha but transparency).
See https://github.com/adobe/asset-compute-sdk/blob/d5095ee4d42296f209b0264228dad6e5e7a575d0/lib/postprocessing/image.js#L54. I'll deal with merging the logic when time to merge https://github.com/adobe/asset-compute-sdk/pull/76 comes.

lib/postprocessing/image.js
```diff
@@ -39,16 +39,23 @@ const SUPPORTED_OUTPUT_FMT = {
...
+const NON_ALPHA_FMT = [
```
tmathern, line 55: FYI - No need to update yet:

In another PR, I named it `FORMAT_SUPPORTS_TRANSPARENCY` (because when people wrote about TIFF, they paid attention to not name it alpha but transparency).
See https://github.com/adobe/asset-compute-sdk/blob/d5095ee4d42296f209b0264228dad6e5e7a575d0/lib/postprocessing/image.js#L54. I'll deal with merging the logic when time to merge https://github.com/adobe/asset-compute-sdk/pull/76 comes.

tmathern: (The way the object is created in https://github.com/adobe/asset-compute-sdk/pull/76 is sorted of a trick for easy and fast hashtable lookup).

lib/postprocessing/image.js
```diff
@@ -459,6 +473,29 @@ async function imagePostProcess(intermediateRendition, rendition, directories) {
...
+    if (NON_ALPHA_FMT.includes(fmt)) {
+        img.background("rgb(255,255,255)"); // Change transparency to white
+        img.flatten(); // forces the image on its background
+    } else {
+        img.background("rgba(0, 0, 0, 0.0)"); // Keep transparency
+    }
+
+    // some svgs have no size (only percentage width/height), so we scale them to our target size
+    img.in("-size", `${width}`)
+    img.colorspace('RGB');
```
tmathern, line 495: I'm curious - why do we need colorspace here?

test-worker/test/asset-compute/worker/svg-png-missing-width/params.json
```diff
@@ -0,0 +1,4 @@
+{
+        "height":100,
```
tmathern, line 2: Indentation is off...

tmathern (approved):

lib/postprocessing/image.js
DrGirlfriend: I think it was from a previous Alex comment ... but now I can't seem to find it
I'll take it out

lib/postprocessing/image.js
```diff
@@ -39,16 +39,23 @@ const SUPPORTED_OUTPUT_FMT = {
...
 const SUPPORTED_INPUT_TYPES = {
     "TIFF": "tif",  // most preferred format as it allows to preserve many features
-    "PNG":  "png",  // lossless over
+    "PNG": "png",  // lossless over
     "JPEG": "jpg",  // ..lossy
-    "GIF":  "gif",  // (should never get here in practice)
-    "BMP":  "bmp"
+    "GIF": "gif",  // (should never get here in practice)
+    "BMP": "bmp",
+    "SVG": "svg"
```
alexkli, line 46: Would be nice to keep the "column/table" style indentation I added for readability.

lib/postprocessing/image.js
```diff
@@ -459,6 +473,28 @@ async function imagePostProcess(intermediateRendition, rendition, directories) {
...
+
+    if (NON_ALPHA_FMT.includes(fmt)) {
+        img.background("rgb(255,255,255)"); // Change transparency to white
+        img.flatten(); // forces the image on its background
+    } else {
+        img.background("rgba(0, 0, 0, 0.0)"); // Keep transparency
+    }
+
+    // some svgs have no size (only percentage width/height), so we scale them to our target size
+    img.in("-size", `${width}`);
```
alexkli, line 494: AFAICS [`img.rawSize()`](http://aheckmann.github.io/gm/docs.html#rawSize) maps to the underlying `-size` flag and we should use that. I assume just setting the first argument (width) should work...

lib/postprocessing/image.js
```diff
@@ -459,6 +473,28 @@ async function imagePostProcess(intermediateRendition, rendition, directories) {
...
 }
 
+function convertSvg(img, instructions) {
+
+    // Note RSVG is the converter here currently 2020/10/22 causing differences between normal Imagemagic use
+    // Future - allow an instruction to define what color should be converted to transparent
+
+    // ImageMagick automatically keeps aspect ratio
+    // Only using width because ImageMagick will use the smallest value whether it be width or height
+    const width = instructions.width || 400;
```
alexkli, line 483: Maybe turn 400 into a constant.

alexkli (approved):
Please add a test with svg as input that specifies no width or height in the input.

(Could be that I missed it)

lib/postprocessing/image.js
tmathern: (done)

lib/postprocessing/image.js
DrGirlfriend: That is what I had previously but if I use img.rawSize(), this will not be applied at the correct time and the SVG will be upscaled resulting in a fuzzy final rendition.

alexkli: If `rawSize()` doesn't add the `-size` in the right location we shouldn't use it. Probably useful to add a comment regarding this.

PR #71, opened 2020-10-19
NUI-855 - Better metrics and logs for post processing

alexkli:
For [NUI-855](https://jira.corp.adobe.com/browse/NUI-855). I slightly changed that from just `Log the duration of just the imagemagick convert command in post processing` to `Better metrics and logs for post processing` and doing all these small improvements:

- Log the duration of the imagemagick convert command in post processing
- Log the duration of the imagemagick metadata read command in post processing
- Add `imagePostProcess` (boolean) metric if image post processing ran
- Change worker timeout message wording to include fmt not worker name

## Types of changes

### DIFF
--- a/lib/worker.js
+++ b/lib/worker.js
@@ -308,6 +308,7 @@ class AssetComputeWorker {
 
         try {
             if (await this.shouldPostProcess(rendition)) {
+                this.metrics.add({ imagePostProcess: true });
 
                 // at this point, we have the rendition a worker created, available at rendition.path
                 // naming rules are rendition0.extension, rendition1.extension, etc.
@@ -332,6 +333,7 @@ class AssetComputeWorker {
 
             } else {
                 this.timers.postProcessing.stop();
+                this.metrics.add({ imagePostProcess: false });
             }
 
             return rendition;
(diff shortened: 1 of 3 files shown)


lib/worker.js
```diff
@@ -308,6 +308,7 @@ class AssetComputeWorker {
 
         try {
             if (await this.shouldPostProcess(rendition)) {
+                this.metrics.add({ imagePostProcess: true });
```
tmathern, line 311: I thought we would also add the post-processing duration here?

tmathern (approved):

lib/worker.js
alexkli: We already have a post processing duration in the metrics, which includes `shouldPostProcess` (e.g. read intermediate metadata), any extra stuff like downloading the watermark and the final imagemagick convert command.

I think that's enough for the metrics, more fine grained measurements here would not be regularly looked at anyway. We only wanted more details in the logs for individual issue analysis (where it came up).

This flag was added so we can query what amount of e.g. worker x runs are with or without post processing.

alexkli:
Query to see the new `imagePostProcess` metric in action on the IT tests:
```sql
SELECT filter(count(*), WHERE imagePostProcess=1) AS 'post-process', filter(count(*), WHERE imagePostProcess=0) AS 'direct' FROM rendition WHERE package='it-20201019-adobe-asset-compute-sdk-e6898dcdff29' FACET actionName, fmt
```

Bildschirmfoto 2020-10-19 um Mo , 19  Okt , 16 47 33

PR #78, opened 2020-11-09
NUI-844 Replace imagemagick metadata reader with exiftool

tmathern:
## Description
- Updates the SDK to rely on exiftool for metadata reading rather than imagemagick.
- Runs an exiftool command, so does not rely on external node libraries (exiftool must be installed in the docker image, though).

Fixes [NUI-844](https://jira.corp.adobe.com/browse/NUI-844).

## Types of changes

## Checklist:


lib/postprocessing/image.js
```diff
@@ -72,8 +72,6 @@ const EXTRA_INSTRUCTIONS = new Set([
     "watermark"
 ]);
 
-const CHANGE_ORIENTATION_LIST = [ "LeftTop", "RightTop", "RightBottom",  "LeftBottom"];
```
tmathern, line 75: With the `-n` option, exiftool returns the numerical value. So it becomes a simpler comparison.

lib/postprocessing/image.js
```diff
@@ -87,23 +85,49 @@ function hasExtraInstructions(instructions) {
...
+            "-FileType",
+            "-ImageHeight",
+            "-ImageWidth",
+            "-JPEGQualityEstimate"
+        ];
+        const attributesToExtract = attributes.join(" ");
+        const command = `exiftool -n -json ${attributesToExtract} ${file}`;
+        console.log(`Metadata (exiftool) command is: ${command}`);
+
+        const { stdout, stderr } = await exec(command);
```
tmathern, line 101: Should I log timing? I don't think it's necessary for exiftool, but open for debate.

lib/postprocessing/image.js
```diff
@@ -87,23 +85,49 @@ function hasExtraInstructions(instructions) {
...
+        if(stderr){
+            throw new GenericError(`Reading metadata from intermediate rendition failed: ${stderr}`);
+        }
 
-        // newer imagemagick returns an array (for multi-page formats), older just one object (6.8.9-9 at least)
-        if (Array.isArray(json)) {
-            return json[0].image;
-        } else {
-            return json.image;
+        let metadata = JSON.parse(stdout);
```
tmathern, line 107: The value returned is a string, not a parsed json object.

lib/postprocessing/image.js
```diff
@@ -87,23 +85,49 @@ function hasExtraInstructions(instructions) {
 }
 
 async function getImageMetadata(file) {
-    const start = new Date();
     try {
-        // [1x1+0+0] is a trick to avoid reading all pixels into memory, improving time for large images
-        const result = await gm(`${file}[1x1+0+0]`).write("json:-");
-        const json = JSON.parse(result.toString());
+        // suffix attributes with `#` to disable human-readable parsing if exiftool not used with `-n` flag
+        const attributes = [
```
tmathern, line 90: I only extract the metadata we need for post-processing handling.

alexkli: I would just extract all metadata. As the file is locally available here, there is not much to win by limiting it, and it probably reads everything anyway.

Having all metadata makes future changes that need more metadata fields easier.

lib/postprocessing/image.js
```diff
@@ -87,23 +85,49 @@ function hasExtraInstructions(instructions) {
...
+            "-JPEGQualityEstimate"
+        ];
+        const attributesToExtract = attributes.join(" ");
+        const command = `exiftool -n -json ${attributesToExtract} ${file}`;
+        console.log(`Metadata (exiftool) command is: ${command}`);
+
+        const { stdout, stderr } = await exec(command);
+
+        if(stderr){
+            throw new GenericError(`Reading metadata from intermediate rendition failed: ${stderr}`);
```
alexkli, line 104: Is it true that if it prints to `stderr` it failed? That would be very unusual. A non-zero exit code should be the trigger.

lib/postprocessing/image.js
```diff
@@ -87,23 +85,49 @@ function hasExtraInstructions(instructions) {
...
+        }
+
+        // normalize to have nicer names (and same names as for imagemagick's metadata reader)
+        const filteredMetadata = {};
+        filteredMetadata.format = metadata.FileType;
+        filteredMetadata.pageGeometry = {};
+        filteredMetadata.pageGeometry.height = metadata.ImageHeight;
+        filteredMetadata.pageGeometry.width = metadata.ImageWidth;
+        filteredMetadata.orientation = metadata.Orientation;
+        filteredMetadata.quality = metadata.JPEGQualityEstimate;
```
alexkli, line 123: I guess we want to change all usages to use the exiftool json output. The one from imagemagick is entirely undocumented, so future maintainers of this code (including us) would have a hard time figuring out what this intermediate object structure (`filteredMetadata`) should be.

lib/postprocessing/image.js
```diff
@@ -225,10 +249,11 @@ async function needsImagePostProcess(rendition, source) {
         return false;
     }
 
-    await logImageMagickVersion();
-
     // inspect intermediate rendition
-    const metadata = rendition._intermediateMetadata = await getImageMetadata(rendition.path);
+    rendition._intermediateMetadata = await getImageMetadata(rendition.path);
+    const metadata = rendition._intermediateMetadata;
```
alexkli, line 254: Well at least the other way around :-)

lib/postprocessing/image.js
tmathern: I didn't manage to make it fail yet - so I removed this.

lib/postprocessing/image.js
tmathern: We don't want that. To get the quality estimate, I would have to force the tool to request all metadata - those stored but also those which must be calculated, like the quality. Also, this forces the tool to generate fields that would not normally be generated (calculated fields, nested fields, parsing of fields stored in hexa). This incurs a performance hit. Also, the metadata object ends up being way larger than what we had before. 

So here I limited it to fields we use already on purpose, to mitigate this. Having `JPEGQualityEstimate` explicitly listed as field here will force its parsing, and only its parsing - so I don't need to force exiftool to parse all metadata and the object we carry around stays smaller in size.

tmathern: `JPEGQualityEstimate ` is an extra tag, see https://exiftool.org/TagNames/Extra.html.

To get all metadata, we would need the `RequestAll` flag set to 3. Again, this value must set to 3 in order to grab the quality estimate. From [documentation](https://exiftool.org/ExifTool.html): Flag to request all tags to be extracted. This causes some tags to be generated which normally would not be unless specifically requested (by passing the tag name to ImageInfo or ExtractInfo). May be set to 2 or 3 to enable generation of some additional tags as mentioned in the tag name documentation.

lib/postprocessing/image.js
tmathern: I would prefer not to. I change `.pageGeometry.height` to `.imageHeight` and `.pageGeometry.width` to `.imageWidth`. But I'd appreciate we don't change the naming depending on every tool we use - hence that very small and basic abstraction.
Also, the object structure is kept flat on purpose.

lib/postprocessing/image.js
DrGirlfriend: Good catch!

DrGirlfriend (approved):
LGTM

lib/postprocessing/image.js
alexkli: 1. you already ask for the `JPEGQualityEstimate` flag - how would asking for all metadata change that part?
2. IIUC with imagemagick this quality was read from a metadata that is usually added to the file. We should definitely avoid calculating it, that doesn't really bring any benefit. Maybe there is an option to tell exiftool to not do that? That would definitely be preferred to keep overall behavior as it is today.
3. As mentioned in the Slack discussion, maybe we could drop [this check](https://github.com/adobe/asset-compute-sdk/blob/65541a51b2df1fd63740c71b877ae2d477ad8f44/lib/postprocessing/image.js#L175) altogether if we think quality is not a reliable metadata. And just assume if it's jpeg, then whoever created it did respect the `instructions.quality`. Just need to make sure this is true if the source file is jpeg already. (To be clear: I forgot that we check for the quality metadata here.).

lib/postprocessing/image.js
alexkli: Please use the exiftool format. That is documented or at least clear. The new one you added is custom and has no documentation. So when new metadata fields are added folks have to invent new stuff and will get confused. Just extracting all metadata and using the exiftool format is good. We do expect to stick with exiftool now :-)

lib/postprocessing/image.js
tmathern: 1. Because `JPEGQualityEstimate` is named explicitly, only this flag is calculated (again, it is *calculated*, not read). Asking for all would trigger calculation for all calculated flags. 
2. No, the hypothesis is wrong. Imagemagick computed it. This depends on the tool that created the file. The metadata will be written while compressing, but removed afterwards usually. This is part of to the [JFIF (JPEG) spec](https://www.w3.org/Graphics/JPEG/jfif3.pdf): **the quality number (0-100) is not stored in the image header**. For other formats see [imagemagick's quality param](https://imagemagick.org/script/command-line-options.php#quality). It always depends on compression, and is never written as a metadata. Photoshop writes also a quality value, which has nothing to do with that one (it's another scale, max value is 12).
3. Who should make the call to drop it?

lib/postprocessing/image.js
tmathern: Just extracting all metadata and using the exiftool format is good. <= No, for reasons explained in https://github.com/adobe/asset-compute-sdk/pull/78#discussion_r520676816, https://github.com/adobe/asset-compute-sdk/pull/78#discussion_r520678938 and https://github.com/adobe/asset-compute-sdk/pull/78#discussion_r520883756. 

Fine for naming. It's a bit a leaky abstraction, but I don't feel like fighting naming here.

lib/postprocessing/image.js
alexkli: > Who should make the call to drop it?

Analyze the situation. Do we have any cases where this could be an issue if we drop?

tmathern: The problem I have with the jpeg quality value is that it's useless at best, and misguiding at worst. The JPEG standard does not actually define the "quality" factor, only how the decompression works. The Discrete Cosine Transformation is usually what's used here, but either through implementation details, or through actual algorithmic differences (see https://en.wikipedia.org/wiki/JPEG#JPEG_compression).

In other words, the quality factor is going to have different meanings for different encoders. Different encoders will have different scales and different meanings for these scales, which causes confusion even for the end-users (see https://photo.stackexchange.com/questions/11544/why-does-jpeg-quality-go-up-to-twelve - Photoshop uses a scale from 0 to 12, whereas other encoders have a scale from 0 to 100).

So if a developer wants to make any actual use of these values, they'll need to use the same exact encoder that was used originally with this quality value. If they use a jpeg that was encoded on a scale from 0 to 12, and pass whichever value from it into an encoder that's working on a 0-100 scale, the results won't be what they expect at all. And of course, the encoder originally used isn't part of the metadata.

Long story short, I strongly believe that having the quality metadata exposed can cause more harm than good.

lib/postprocessing/image.js
```diff
@@ -12,13 +12,14 @@
...
 
-const { OpenwhiskActionName, GenericError, RenditionFormatUnsupportedError } = require('@adobe/asset-compute-commons');
+const { OpenwhiskActionName, GenericError, RenditionFormatUnsupportedError, SourceCorruptError } = require('@adobe/asset-compute-commons');
 
 const util = require('util');
 const exec = util.promisify(require('child_process').exec);
 
 const { getAsset } = require('../storage');
 const { Dimensions } = require("./math");
+const unitConvert = require('css-unit-converter');
```
tmathern, line 22: Sizing information, if present in an SVG, may have units. And this unit may be something else than `pixels`, so we would need to convert. This module is for doing conversions.

lib/postprocessing/image.js
```diff
@@ -288,14 +355,14 @@ async function read(infile) {
...
 
 function getImageSize(metadata) {
-    // The orientation corresponds to the TIFF orientation tag which ranges fromn 1 to 8
+    // The orientation corresponds to the TIFF orientation tag which ranges from 1 to 8
     // Values 5 to 8 require a swap of width and height
     // See https://www.awaresystems.be/imaging/tiff/tifftags/orientation.html or
     // https://www.adobe.io/content/dam/udp/en/open/standards/tiff/TIFF6.pdf
-    if (CHANGE_ORIENTATION_LIST.includes(metadata.orientation)) {
-        return new Dimensions(metadata.pageGeometry.height, metadata.pageGeometry.width);
+    if (parseInt(metadata.Orientation, 10) >= 5) {
```
tmathern, line 362: Exiftool gives out the numerical value (`-n` option).

lib/postprocessing/image.js
```diff
@@ -233,10 +294,16 @@ async function needsImagePostProcess(rendition, source) {
...
-    await logImageMagickVersion();
-
     // inspect intermediate rendition
-    const metadata = rendition._intermediateMetadata = await getImageMetadata(rendition.path);
+    const metadata = await getImageMetadata(rendition.path);
+    rendition._intermediateMetadata = metadata;
+
+    if(!metadata.FileType){
+        // could not recognize filetype - most likely it is corrupt
+        throw new SourceCorruptError(`Rendition seems to be corrupt. Could not identify file type.`);
```
tmathern, line 303: We can now detect additional cases in which the source file seems to be corrupt.

package.json
```diff
@@ -19,6 +19,7 @@
     "ajv": "6.12.3",
     "clone": "^2.1.2",
     "content-type": "^1.0.4",
+    "css-unit-converter": "^1.1.2",
```
tmathern, line 22: Best library I could find for easy sizing unit conversions (needed because exiftool gives out sizes with units, which are not always necessarily pixels!).

PR #87, opened 2020-11-18
NUI-916 Use imagemagick as fallback to read metadata when exiftool can't

tmathern:
## Description

Exiftool doesn't known filetypes SGI, RGB or RGBA (for instance), and can't read their metadata. This prevents proper rendering of the files... 

Yet, imagemagick can handle those files. Therefore, when exiftool can't read metadata, we should fallback to using imagemagick. Since this should be pretty rare (currently only with RGB, RGBA and SGI files), I think we should not have too many performance issues related to file sizes.

This will help us to support more file formats (especially those not recognized by exiftool).

Fixes #[NUI-916](https://jira.corp.adobe.com/browse/NUI-916)

## Types of changes

## Checklist:


jdelbick:
question for optimization because (as I have noticed), exiftool is heavy on memory and some of our actions have the default 256MB (I think, maybe this is false):

> Exiftool doesn't known filetypes SGI, RGB or RGBA (for instance), and can't read their metadata.
Do these file types never work with exiftool? If so I recommend not even trying exiftool with them and going straight to imagemagick.

another question: Which is faster/more efficient: exiftool or imagemagick? do we want to reverse it to try imagemagick then fallback on exiftool? (if exiftool is faster for all file types, ignore this)

jdelbick (approved):

jdelbick:
btw approving, we can make a ticket to optimize this in the future

tmathern:
`Do these file types never work with exiftool?` => SGI, RGB amd RGBA never work with exiftool. They are currently unsupported. 

`Do these file types never work with exiftool? If so I recommend not even trying exiftool with them and going straight to imagemagick.` => I have sort of a chicken-egg problem: because I'm trying to get filetypes from metadata rather than the extension, so I can't immediately direct them to imagemagick (except by parsing the extension, but I'd like to avoid that, as we can't really trust extensions...). Also, maybe someone one day will add support for those in exiftool (you never know). 

`exiftool is heavy on memory and some of our actions have the default 256MB` => Not for workers that do image processing. They always have more (smallest is 512Mb I think - and this is one which only extracts thumbnails from files). And those workers which do image processing and need post-processing usually have even more memory (1Gb at least). Other workers with small memory settings (zipping, for instance), do not do image renditions and will never go in the codepath where calling exiftool is needed. So we're good on that front too!

To finish, I'm not sure which one between imagemagick and exiftool is faster, but exiftool is more memory efficient than imagemagick for larger files.

jdelbick:
@tmathern since most actions have high memory, I say we only optimize (between imagemagick & exiftool based on file ext) if it comes up as an issue

tmathern:
We would have to always double-check anyway that we have the correct file extension. Nothing really prevents you, for instance, to turn a `.tiff` into a `.jpeg` just by changing the file name. 
(We had issues because of that already before).

Also, first let's have something work correctly, as expected without (too many) surprises, then optimize ;) !

DrGirlfriend (approved):

PR #103, opened 2021-04-20
fix failing unit tests, file command

jdelbick:
## Description

newer `file` command now returns `image/bmp` for the bmp file. I even tried it locally:
```
$ file -b --mime /Users/jamiedelbick/work/adobe/asset-compute-sdk/test/files/file.bmp                     
image/bmp; charset=binary
``` 

Fixes # (issue)

## Types of changes

## Checklist:

### DIFF
--- a/test/rendition.test.js
+++ b/test/rendition.test.js
@@ -367,7 +367,7 @@ describe("rendition.js", () => {
             // overwrite path to point to test files
             rendition.path = './test/files/file.bmp';
             const result = await rendition.mimeType();
-            assert.strictEqual(result, 'image/bmp');
+            assert.ok(result === 'image/bmp' || result === 'image/x-ms-bmp');
         });
 
         it('gracefully handles not finding files when identifying mimetype', async function () {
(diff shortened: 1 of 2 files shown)


test/rendition.test.js
```diff
@@ -367,7 +367,7 @@ describe("rendition.js", () => {
             // overwrite path to point to test files
             rendition.path = './test/files/file.bmp';
             const result = await rendition.mimeType();
-            assert.strictEqual(result, 'image/x-ms-bmp');
+            assert.strictEqual(result, 'image/bmp');
```
tmathern, line 370: I'd cover this case this way, actually:

```suggestion
            assert.ok(result === 'image/bmp' || result === 'image/x-ms-bmp');
```

Wdyt?

pheenomenon (approved):
Was it the file command in our base image which was updated and might have caused this ?
https://git.corp.adobe.com/nui/openwhisk-runtime-nodejs/blob/master/core/nodejs10Action-centos/Dockerfile#L153

jdelbick:
@pheenomenon, no this happened in the mocha unit tests (not the test-worker tests) that are not running inside one of our containers. this means `file` is whatever version is installed on this: https://github.com/adobe/asset-compute-sdk/blob/master/.github/workflows/node.js.yml#L15

badvision (approved):
This is fine.  If we have to support additional mime types for some reason then we might at some point want to represent them as an array and use something like Array.includes against that list of mime types.  For only two types there's no real gain from it though.

PR #106, opened 2021-05-06
Repro CI/CD bug

tmathern:
## Description

Fixes # (issue)

## Types of changes

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
   "dependencies": {
     "@adobe/asset-compute-commons": "^1.1.2",
     "@adobe/cgroup-metrics": "^3.0.3",
-    "@adobe/httptransfer": "^2.1.1",
+    "@adobe/httptransfer": "^2.2.0",
     "@adobe/metrics-sampler": "^1.0.2",
     "@adobe/node-fetch-retry": "^1.1.1",
     "ajv": "6.12.3",
--- a/package-lock.json
+++ b/package-lock.json
@@ -323,14 +323,17 @@
       }
     },
     "@adobe/httptransfer": {
-      "version": "2.1.1",
-      "resolved": "https://registry.npmjs.org/@adobe/httptransfer/-/httptransfer-2.1.1.tgz",
-      "integrity": "sha512-c2lZdN84LZW5mOC9cPVRqdkoVlOhH35snYLk9HaOLa8DqBuXztSGr4P9JGuCkzmSCPbqZImYR89wSliHVF6Nyw==",
+      "version": "2.2.0",
+      "resolved": "https://registry.npmjs.org/@adobe/httptransfer/-/httptransfer-2.2.0.tgz",
+      "integrity": "sha512-1ykNA7wpC8i5yT0fAdYtT+hcU8+mo6+H+xobkucxlA5z/aiATFACk0P4+NvpnQNhIoCM2r4arVWNXb55DPZ9NQ==",
       "requires": {
+        "@babel/runtime": "^7.14.0",
         "content-disposition": "^0.5.3",
         "content-range": "^2.0.0",
         "content-type": "^1.0.4",
+        "core-js": "^3.11.3",
         "drange": "^2.0.1",
+        "file-url": "2.0.2",
         "filter-obj": "^2.0.1",
         "fs-extra": "^9.1.0",
         "node-fetch-npm": "^2.0.4"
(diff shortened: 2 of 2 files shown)


test-worker/manifest.yml
```diff
@@ -5,3 +5,6 @@ packages:
       worker:
         function: worker.js
         docker: 'alexkli/nodejs10-action-centos:2.4.1'
+      limits:
+        memorySize: 4096
```
tmathern, line 9: bmp with transparency -> gif (which supports transparency) seems to be more memory intensive than I'd had expected

test/postprocess.test.js
```diff
@@ -115,23 +117,28 @@ describe("postprocessing/image.js", () => {
...
-        assert.equal(receivedMetrics[1].callbackProcessingDuration + receivedMetrics[1].postProcessingDuration, receivedMetrics[1].processingDuration);
-        assert.equal(receivedMetrics[0].imagePostProcess, true);
-        assert.equal(receivedMetrics[1].imagePostProcess, true);
+
+        assert.strictEqual(receivedMetrics[0].eventType, "rendition");
+        assert.strictEqual(receivedMetrics[0].callbackProcessingDuration + receivedMetrics[0].postProcessingDuration, receivedMetrics[0].processingDuration);
+        assert.strictEqual(receivedMetrics[1].eventType, "activation");
+        assert.strictEqual(receivedMetrics[1].callbackProcessingDuration + receivedMetrics[1].postProcessingDuration, receivedMetrics[1].processingDuration);
+
+        // queued metrics may "flatten metrics" and turn booleans true/false into 1/0
```
jdelbick, line 139: interesting find. I think we should always flatten the metrics regardless of if they are sent in the queue or not? wdyt?

looks like they may always flatten actually since they are always sent through the queue class: https://github.com/adobe/node-openwhisk-newrelic/blob/master/lib/newrelic.js#L151

jdelbick (approved):
LGTM

test/postprocess.test.js
tmathern: Yeah. I think mostly the tests didn't pick up the change. But to be sure, I updated the tests to accept both.

package.json
```diff
@@ -13,7 +13,7 @@
   "dependencies": {
     "@adobe/asset-compute-commons": "^1.1.2",
     "@adobe/cgroup-metrics": "^3.0.3",
-    "@adobe/httptransfer": "^2.1.1",
+    "@adobe/httptransfer": "^2.2.0",
```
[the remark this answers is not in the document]
tmathern: It will bring in his changes. But so would any reinstall of the SDK (in any place it's used), since the caret `^` means to grab any bugfix or feature update. I'm just making it explicit now. If we have to keep the version without changes, we would need to pin the version, or use the tilde `~` instead (tilde only grabs patches/bugfixes).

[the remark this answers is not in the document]
tmathern: Also, all unit and worker tests pass, so it looks pretty safe to get it in.

tmathern:
Added a set of tests to check colorspace handling. The used source file has the particularity of being able to give widely different renditions if colorspace are not handled as we want them to (namely, the strength of lights will vary strongly, has will the darkest spots).

package.json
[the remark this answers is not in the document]
tmathern: @bobvanmanen Wdyt?

test-worker/test/asset-compute/worker/validate-image.sh
```diff
@@ -35,7 +35,7 @@ function im_identify() {
 
 # check for pixel equality with some slight accepted color value difference
 # will exit with 1 if images differ
-FUZZ="5%"
+FUZZ="10%"
```
pheenomenon, line 38: why increase FUZZ ? AFAICS, This isn't used in colorspace validation

tmathern:
https://github.com/adobe/asset-compute-sdk/pull/110

pheenomenon (approved):

test-worker/test/asset-compute/worker/validate-image.sh
tmathern: Fuzz got increase to not flag tests as failed when renditions got created by 2 different imagemagick versions (note that in other tests it's up to 15%). And the fuzz value is used - the "normal" validation script (this one) gets called after any specific/custom validation.

pheenomenon: Ok thanks!
I know one imagemagick is the one used inside https://github.com/adobe/asset-compute-sdk/blob/master/test-worker/manifest.yml#L7 , which is the other one(for this scenario)?

tmathern: The one from our latest base image, 7.0.11-6 (with latest supporting libraries too).

PR #114, opened 2021-06-03
NUI-1275 implementation manifest to its own file

pheenomenon:
For https://jira.corp.adobe.com/browse/NUI-1275 [design decision + impl] decide how to import multiple manifest files into workers

Based on our discussion at the end of the mob(with Alex), reverting some of our changes from our mob yesterday.
- Make manifest a separate file.

## Description

Fixes # (issue)

## Types of changes

## Checklist:

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -151,16 +151,13 @@ function shellScriptWorker(script = "worker.sh", options = {}) {
 }
 
 async function buildTransformers(renditionCallback, options={}) {
-    const transformers = [];
     if (!options.manifest) {
-        return transformers;
+        return [];
     }
-    
     debug('Manifest', options.manifest);
     const transformer = new WorkerCallbackTransformer(renditionCallback, options.manifest);
     debug(`Created a transformer named ${transformer.name}`);
-    transformers.push(transformer);
-    return transformers.concat(options.transformerCatalog);
+    return [transformer, ...options.transformerCatalog];
 }
 
 module.exports = {
--- a/lib/api.js
+++ b/lib/api.js
@@ -152,22 +152,15 @@ function shellScriptWorker(script = "worker.sh", options = {}) {
 
 async function buildTransformers(renditionCallback, options={}) {
     const transformers = [];
-
-    if (!options.manifests) {
+    if (!options.manifest) {
         return transformers;
     }
-
-    const manifests = options.manifests;
-    console.log('Manifest array', manifests);
-
-    manifests.forEach(oneManifest => {
-        const oneTransformer = new WorkerCallbackTransformer(renditionCallback, oneManifest);
-        debug(`Created a transformer named ${oneTransformer.name} with manifest:`);
-        debug(oneManifest);
-        transformers.push(oneTransformer);
-    });
-    return transformers.concat(options.transformerCatalog);
     
+    debug('Manifest', options.manifest);
+    const transformer = new WorkerCallbackTransformer(renditionCallback, options.manifest);
+    debug(`Created a transformer named ${transformer.name}`);
+    transformers.push(transformer);
+    return transformers.concat(options.transformerCatalog);
 }
 
 module.exports = {
--- a/test-worker/worker.js
+++ b/test-worker/worker.js
@@ -16,7 +16,6 @@
 
 const { worker, GenericError } = require('../index');
 const gm = require("../lib/postprocessing/gm-promisify");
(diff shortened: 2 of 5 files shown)


lib/api.js
```diff
@@ -152,22 +152,15 @@ function shellScriptWorker(script = "worker.sh", options = {}) {
...
-
-    manifests.forEach(oneManifest => {
-        const oneTransformer = new WorkerCallbackTransformer(renditionCallback, oneManifest);
-        debug(`Created a transformer named ${oneTransformer.name} with manifest:`);
-        debug(oneManifest);
-        transformers.push(oneTransformer);
-    });
-    return transformers.concat(options.transformerCatalog);
     
+    debug('Manifest', options.manifest);
```
tmathern, line 159: Add the transformer name to the log line, so it's easy to debug (in case it crashes during the constructor call below)
```suggestion
    debug(`Manifest for transformer ${transformer.name}`, options.manifest);
```

lib/api.js
```diff
@@ -152,22 +152,15 @@ function shellScriptWorker(script = "worker.sh", options = {}) {
...
+    const transformer = new WorkerCallbackTransformer(renditionCallback, options.manifest);
+    debug(`Created a transformer named ${transformer.name}`);
+    transformers.push(transformer);
+    return transformers.concat(options.transformerCatalog);
```
tmathern, line 163: Alternative:
```suggestion
    return transformers.concat(transformer, options.transformerCatalog);
```
(concat can take many params)

lib/api.js
```diff
@@ -152,22 +152,15 @@ function shellScriptWorker(script = "worker.sh", options = {}) {
...
-        debug(`Created a transformer named ${oneTransformer.name} with manifest:`);
-        debug(oneManifest);
-        transformers.push(oneTransformer);
-    });
-    return transformers.concat(options.transformerCatalog);
     
+    debug('Manifest', options.manifest);
+    const transformer = new WorkerCallbackTransformer(renditionCallback, options.manifest);
+    debug(`Created a transformer named ${transformer.name}`);
+    transformers.push(transformer);
```
tmathern, line 162: Drop this line, to do a one-liner on the return statement... 
```suggestion
```

test-worker/worker.js
```diff
@@ -70,15 +69,5 @@ exports.main = worker(async (source, rendition) => {
...
-            "name": "sdkTestWorker",
-            "inputs": {
-                "type": ["image/jpeg", "image/png"]
-            },
-            "outputs": {
-                "type": ["image/jpeg", "image/png"]
-            }
-        })
-    ]
+    manifest: require("./pipeline-manifest.json")
```
tmathern, line 72: Ok for now - later we'll need validation.
(And a test case to verify behavior when the file is missing, not acceptable etc).

tmathern (approved):
Some suggestions, but already approving (lgtm).

jdelbick (approved):
I think tania's requested changes accomplish a similar simplification that I was thinking of, but we shouldn't need to declare a transformers array: https://github.com/adobe/asset-compute-sdk/blob/22e4f1fec8d00e3d2a3b9169d629006884d4d4f7/lib/api.js#L154
we can just do this:
```
async function buildTransformers(renditionCallback, options={}) {     
if (!options.manifest) {        return [];    }        
debug('Manifest', options.manifest);    
const transformer = new WorkerCallbackTransformer(renditionCallback, options.manifest);    
debug(`Created a transformer named ${transformer.name}`);    
transformers.push(transformer);    
return  [transformer, ...transformersCatalog];}
```

lib/api.js
pheenomenon: got that at line 161.

lib/api.js
pheenomenon: got that at line 161

test-worker/worker.js
pheenomenon: Added a test(currently skipped), but not getting `rewire` to work (because of npm link asset-compute-pipeline, I think) 
Will come back to it when we release pipeline, unless if there is any other way to achieve rewiring..

lib/api.js
tmathern: I know, but if constructing the object fails, we won't have the transformer name logged. My request is to long earlier to make sure we have it.

pheenomenon: we don't have `transformer` until line 161. If `WorkerCallbackTransformer` constructor fails, we have the manifest logged.

PR #120, opened 2021-06-15
Pipeline batch worker support

tmathern:
## Description

Fixes # (issue)

## Types of changes

## Checklist:


lib/worker-pipeline.js
```diff
@@ -112,25 +112,101 @@ class AssetComputeWorkerPipeline {
...
-        debug("Refined plan to create rendition");
+        debug("compute: Refined plan to create rendition: ", plan.toString());
 
-        debug("Running pipeline...");
+        debug("compute: Running pipeline...");
         const pipelineOutput = await engine.run(plan);
-        debug("Pipeline finished running");
-        debug("######## pipelineOutput:", pipelineOutput);
+        debug("compute: Pipeline finished running");
+        debug("compute: pipeline output:", pipelineOutput);
```
jdelbick, line 122: note: I noticed some of our workers run in earlier nodejs versions where `debug` does not support comma-separated logs. we should probably stick to template literals for this reason

lib/worker-transformer.js
```diff
@@ -28,10 +28,18 @@ class WorkerCallbackTransformer extends Transformer {
 
     compute(input, output) {
         if(typeof this._callback !== "function"){
-            throw new Error("renditionCallback must be a function");
+            throw new Error("compute: renditionCallback must be a function");
         }
         return this._callback(input, output, input.params);
     }
+
+    computeAllAtOnce(renditionsCallback) {
```
jdelbick, line 36: this function needs to be called `compute` since that is what the transformer api expects.

I would create a new worker-transformer called `batchWorkerTransformer` that has a `compute` function that passes the correct params to the batchWorker cb: https://github.com/adobe/asset-compute-sdk#rendition-callback-for-batchworker-required
something like this:
```js
compute(input, output) {
    const outputs = [output]; // batch worker expects a list of renditions
    const outdir =  output.directory; // batch worker expects an output directory: https://git.corp.adobe.com/nui/asset-compute-pipeline/blob/mob-session/lib/sdk/rendition.js#L66
    return this._callback(input, outputs, outdir, params);
}
```

jdelbick (changes requested):
`workerCallbackTransformer` must only have a `compute` function defined. `computeAllAtOnce` won't get called. This api is stable and won't be changed: https://git.corp.adobe.com/nui/asset-compute-pipeline/blob/mob-session/lib/engine.js#L186 (you can mark my words on that one)

to accomplish this for batchWorker, we probably need a different transformer since the `compute` function will behave differently since batch worker has different params. see my comment regarding this: https://github.com/adobe/asset-compute-sdk/pull/120#discussion_r652873717

PR #121, opened 2021-06-16
NUI-1347 [sdk] map nui fmt to mimetype in source and rendition instructions (pipeline only)

pheenomenon:
For https://jira.corp.adobe.com/browse/NUI-1347 [sdk] map nui fmt to mimetype in source and rendition instructions (pipeline only)

- Adds mapping to populate input.type & output.type for pipeline flow.
- Adds test cases for all pipeline MVP formats/source combinations

## Description

Fixes # (issue)

## Types of changes

## Checklist:

### DIFF
--- a/lib/worker-pipeline.js
+++ b/lib/worker-pipeline.js
@@ -82,7 +82,13 @@ class AssetComputeWorkerPipeline {
         }
 
         // source.mimetype will always exist, no need to cultivate here.
-        input.type = input.mimetype && input.mimetype.toLowerCase();
+        const type = input.mimetype && input.mimetype.toLowerCase();
+        // make source mimetypes align to Manifest
+        const SOURCE_TYPE_MAP = {
+            "image/jpg" : "image/jpeg",
+            "image/tif" : "image/tiff"
+        };
+        input.type = SOURCE_TYPE_MAP[type] || (type && type.toLowerCase());
 
         if(!input.type || !output.type) {
             //TODO should throw ClientError ?
--- a/lib/worker-pipeline.js
+++ b/lib/worker-pipeline.js
@@ -71,9 +71,8 @@ class AssetComputeWorkerPipeline {
     }
 
     normalizeInputOuput(input, output) {
-        console.log('input before normalization: ', input);
-        console.log('outpu before normalization: ', output);
-
+        // TODO: we will have special cases for beta-worker-creative 
+        // special case for sensei
         if(output.fmt === 'machine-metadata-json') {
             output.type = 'machine-metadata-json';
         } else {
@@ -85,10 +84,9 @@ class AssetComputeWorkerPipeline {
         // source.mimetype will always exist, no need to cultivate here.
         input.type = input.mimetype && input.mimetype.toLowerCase();
 
-        console.log('input after normalization: ', input);
-        console.log('outpu after normalization: ', output);
         // TODO: throw if params.source.type || rendition.type does not exist
         if(!input.type || !output.type) {
+            //TODO should throw ClientError ?
             console.log("Pipeline won't be able to find a plan and will throw error");
         }
     }
--- a/lib/worker-pipeline.js
+++ b/lib/worker-pipeline.js
@@ -18,6 +18,7 @@ const WorkerCallbackTransformer = require('./worker-transformer');
 
 const fs = require('fs-extra');
 const path = require('path');
+const mime = require('mime-types');
 
 /**
  * Represents a pipeline build around a worker callback.
@@ -69,6 +70,29 @@ class AssetComputeWorkerPipeline {
         return engine;
     }
 
+    normalizeInputOuput(input, output) {
+        console.log('input before normalization: ', input);
+        console.log('outpu before normalization: ', output);
+
+        if(output.fmt === 'machine-metadata-json') {
+            output.type = 'machine-metadata-json';
+        } else {
+            // rendition.fmt should always exist
+            const mimetype = mime.lookup(output.fmt.toLowerCase());
+            output.type = mimetype && mimetype.toLowerCase();
+        }
+
+        // source.mimetype will always exist, no need to cultivate here.
+        input.type = input.mimetype && input.mimetype.toLowerCase();
+
+        console.log('input after normalization: ', input);
+        console.log('outpu after normalization: ', output);
+        // TODO: throw if params.source.type || rendition.type does not exist
+        if(!input.type || !output.type) {
+            console.log("Pipeline won't be able to find a plan and will throw error");
+        }
+    }
+
     async compute(params){
         const engine = this.prepareEngine(params);
         debug("Created pipeline engine");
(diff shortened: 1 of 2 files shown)


lib/worker-pipeline.js
```diff
@@ -69,6 +70,29 @@ class AssetComputeWorkerPipeline {
...
+            // rendition.fmt should always exist
+            const mimetype = mime.lookup(output.fmt.toLowerCase());
+            output.type = mimetype && mimetype.toLowerCase();
+        }
+
+        // source.mimetype will always exist, no need to cultivate here.
+        input.type = input.mimetype && input.mimetype.toLowerCase();
+
+        console.log('input after normalization: ', input);
+        console.log('outpu after normalization: ', output);
```
tmathern, line 89:
```suggestion
        console.log('output after normalization: ', output);
```

test/worker-pipeline.test.js
```diff
@@ -0,0 +1,228 @@
+/*
+ * Copyright 2020 Adobe. All rights reserved.
```
tmathern, line 2:
```suggestion
 * Copyright 2021 Adobe. All rights reserved.
```

lib/worker-pipeline.js
```diff
@@ -69,6 +70,29 @@ class AssetComputeWorkerPipeline {
         return engine;
     }
 
+    normalizeInputOuput(input, output) {
+        console.log('input before normalization: ', input);
+        console.log('outpu before normalization: ', output);
```
jdelbick, line 75:
```suggestion
        console.log('output before normalization: ', output);
```

lib/worker-pipeline.js
```diff
@@ -69,6 +70,29 @@ class AssetComputeWorkerPipeline {
...
+
+        if(output.fmt === 'machine-metadata-json') {
+            output.type = 'machine-metadata-json';
```
jdelbick, line 78: note: we will have special cases for beta-worker-creative too

lib/worker-pipeline.js
```diff
@@ -69,6 +70,29 @@ class AssetComputeWorkerPipeline {
...
+        } else {
+            // rendition.fmt should always exist
+            const mimetype = mime.lookup(output.fmt.toLowerCase());
```
jdelbick, line 81: does mime.lookup throw if there is no valid mimetype or in any other case?

lib/worker-pipeline.js
```diff
@@ -69,6 +70,29 @@ class AssetComputeWorkerPipeline {
...
+            output.type = mimetype && mimetype.toLowerCase();
+        }
+
+        // source.mimetype will always exist, no need to cultivate here.
+        input.type = input.mimetype && input.mimetype.toLowerCase();
```
jdelbick, line 86: what if source mimetype does `image/jpg` instead of `image/jpeg`? (or is this not a valid mimetype)
same goes for `image/tiff` vs `image/tif`

jdelbick (approved):

lib/worker-pipeline.js
pheenomenon: thanks, will add a TODO

lib/worker-pipeline.js
pheenomenon: Doesn't throw, but returns `false`. Added a test for it.
https://www.npmjs.com/package/mime-types -> `All functions return false if input is invalid or not found.`

lib/worker-pipeline.js
pheenomenon: if not already passed in, source gets it mimetype from request headers(in core). Can Manifest handle these synonyms, if not what does Manifest rely on 3 letter or 4 letter ? Can add a static mapping here too.

pheenomenon: looks like Manifest always look for 4 letter mime types i.e. `image/jpeg` & `image/tiff`

lib/worker-pipeline.js
```diff
@@ -69,6 +70,26 @@ class AssetComputeWorkerPipeline {
         return engine;
     }
 
+    normalizeInputOuput(input, output) {
+        // TODO: we will have special cases for beta-worker-creative 
+        // special case for sensei
+        if(output.fmt === 'machine-metadata-json') {
+            output.type = 'machine-metadata-json';
```
tmathern, line 77: In theory we could make ourselves smarter and investigate if we could use a custom mimetype ([it would contain a vnd keyword](https://en.wikipedia.org/wiki/Media_type#Vendor_tree)).

lib/worker-pipeline.js
tmathern: `image/jpg` is not a valid `jpeg` mimetype (sometimes you'll see it because people build the mimetype using the extension). Same for tiff: `image/tif` is not a valid mimetype.

lib/worker-pipeline.js
```diff
@@ -69,6 +70,26 @@ class AssetComputeWorkerPipeline {
...
+            const mimetype = mime.lookup(output.fmt.toLowerCase());
+            output.type = mimetype && mimetype.toLowerCase();
+        }
+
+        // source.mimetype will always exist, no need to cultivate here.
+        input.type = input.mimetype && input.mimetype.toLowerCase();
+
+        if(!input.type || !output.type) {
+            //TODO should throw ClientError ?
+            console.log("Pipeline won't be able to find a plan and will throw error");
```
tmathern, line 89: For MVP, not having a mimetype would be OK, given the types we support should be found.
So You could throw here.

Longer term: We'll need a fallback to find out mimetypes, possibly using `file`+`imagemagick` (one alone does not cover all input types we claim to support). Feel free to write a JIRA story for this if you feel like it :).

test/worker-pipeline.test.js
```diff
@@ -0,0 +1,241 @@
...
+            target: "https://example.com/target.png",
+            name: "cq.dam.319x319.png",
+            fmt: "png"
+        };
+        
+        const testPipelineWorker = new AssetComputeWorkerPipeline();
+        testPipelineWorker.normalizeInputOuput(input, output);
+        assert.strictEqual(input.type,'image/png');
+        assert.strictEqual(output.type,'image/png');
+    });
```
tmathern, line 41: Nitpick: COuld you add one free line in-between tests?

tmathern (approved):
LGTM for a merge into the mob-session branch.

lib/worker-pipeline.js
```diff
@@ -108,12 +129,15 @@ class AssetComputeWorkerPipeline {
             };
             debug('######### input:', input);
         }
+        const output = params.renditions[0];
+        // normalize source and rendition
+        this.normalizeInputOuput(input, output);
```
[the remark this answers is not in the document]
pheenomenon: good idea, had the same thought. But as the originals are available in `params` didn't bother.

lib/worker-pipeline.js
pheenomenon: got it, to be defensive have added a mapping layer for `tif` & `jpg` for source. We can choose to extend it or drop it later as we integrate.

PR #125, opened 2021-06-17
Add batchworker support to SDK

tmathern:
- Adds batchworker support to SDK


badvision (approved):

lib/batch-worker-transformer.js
```diff
@@ -0,0 +1,50 @@
...
+
+    /**
+     * Construct a transformer for batch workers
+     * @param {*} callback 
+     * @param {*} manifest worker manifest (only one, not a list)
+     */
+    constructor(callback, manifest) {
+        super(`batch-workerCallback-${manifest.name}`, manifest);
+
+        if(this.manifest.inputs.hasBatchMode === true || this.manifest.inputs.hasBatchMode === true){
```
pheenomenon, line 27: is this type to check same condition twice ?

lib/worker-pipeline.js
```diff
@@ -105,9 +115,13 @@ class AssetComputeWorkerPipeline {
             input.type = SOURCE_TYPE_MAP[type] || (type && type.toLowerCase());
         }
 
-        if(!input.type || !output.type) {
-            // plan alg will throw error and send io events
-            console.log("Pipeline won't be able to find a plan and will throw error");
+        if (!input.type) {
+            //TODO should throw ClientError ?
```
jdelbick, line 119: please keep comment:
```
// plan alg will throw error and send io events
```
this is the current behavior

lib/worker-pipeline.js
```diff
@@ -57,7 +58,15 @@ class AssetComputeWorkerPipeline {
             return this.transformers;
         }
 
-        const transformer = new WorkerCallbackTransformer(renditionCallback, this.options.manifest);
+        //const transformer = new WorkerCallbackTransformer(renditionCallback, this.options.manifest);
+        let transformer;
+        if (this.options.manifest.hasBatchMode || this.options.manifest.inputs.hasBatchMode) {
```
jdelbick, line 63: remove the second option: `this.options.manifest.inputs.hasBatchMode`

jdelbick (approved):

lib/batch-worker-transformer.js
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    async compute(input, output) {
+        console.log('BatchWorkerCallbackTransformer#compute#compute');
+        if(typeof this._callback !== "function"){
+            throw new Error("compute: renditionCallback must be a function");
+        }
+
+        // the params change compared to `worker-transformer.js` WorkerCallbackTransformer
+        return this._callback(input, output, output.directory);
```
jdelbick, line 46: I am assuming pie doesnt use the 4th optional param `params`?

jdelbick (approved):

lib/worker-pipeline.js
pheenomenon, line 63: We already know here that the call is in batch mode. As you got here from `batchWorker` call in `api.js` (line 72).
So adding an api/method to `AssetComputeWorkerPipeline` will help you decouple from the `hasBatchMode` check and also you won't need to add this field in Manifest(which it seems does not belong there)

lib/batch-worker-transformer.js
tmathern: Ah typo. I wanted to check 2 locations, manifest root and inputs.

lib/worker-pipeline.js
tmathern: Note that you need to clarify the code if you need a comment to describe the behavior. The pipeline here should not worry about if/why the plan algorithm will do something.
I put back your comment, but please please clarify the behavior. Because if we rely on comments, that is a major readability issue for the code.

lib/worker-pipeline.js
tmathern: When we will have a stable API, yes, sure.

tmathern: I don't totally agree, since the manifest is supposed to describe the worker's capabilities, and "batch" is one of those capabilities.

tmathern:
Merging to get all changes together and continue testing in workers...

PR #132, opened 2021-07-13
[NUI-1431] Do not send invalid/empty values as part of metadata for file width/height

tmathern:
For [NUI-1431](https://jira.corp.adobe.com/browse/NUI-1431).

Invalid metadata values (null, undefined, NaN and +/-Infinite) will be removed from the object to be sent, to prevent client exceptions when parsing them.


lib/postprocessing/image.js
```diff
@@ -328,6 +278,56 @@ async function prepareImagePostProcess(rendition, options) {
     // return undefined if no instruction change needed
 }
 
+async function getImageMetadata(file) {
```
tmathern, line 281: I changes the order in the file so metadata-related functions are together towards the bottom, as there are good chances that `metadata` stuff will go into its own file.

lib/rendition.js
```diff
@@ -236,6 +220,35 @@ class Rendition {
             console.log("could not determine encoding/charset");
         }
 
+        try {
+            meta[Metadata.REPO_SIZE] = this.size();
+        } catch (err) {
+            console.log("could not determine file size:", err);
+        }
+
+        try {
```
tmathern, line 229: Moved it down so I can log mimetype easily too.

lib/rendition.js
```diff
@@ -236,6 +220,35 @@ class Rendition {
...
+            const dimensions = await imageSize(this.path);
```
tmathern, line 230: The library we use here, [image-size]() has some limitations that can lead to width/height not being found:
- some JPEGs are [not supported](https://github.com/image-size/image-size/blob/main/lib/types/jpg.ts#L1-L4);
- bigendian TIFF files [are currently not supported](https://github.com/image-size/image-size/blob/main/lib/types/tiff.ts#L2).

lib/rendition.js
```diff
@@ -236,6 +220,35 @@ class Rendition {
...
+            meta[Metadata.TIFF_IMAGEWIDTH] = dimensions.width;
+            meta[Metadata.TIFF_IMAGEHEIGHT] = dimensions.height;
+        } catch (err) {
+            // The rendition may or may not be an image, so log error for informational purposes
+            console.log(`Error trying to find dimensions, no dimensions found:`, err.message || err);
+        }
+
+        // ensure we don't send null/undefined/non-finite values
+        if(meta[Metadata.TIFF_IMAGEWIDTH] === null
```
tmathern, line 239: In case width is undefined, null or an Infinite (which is an invalid integer value in our case), delete the property to ensure no null value gets serialized/send.

lib/rendition.js
```diff
@@ -236,6 +220,35 @@ class Rendition {
...
+        }
+
+        // ensure we don't send null/undefined/non-finite values
+        if(meta[Metadata.TIFF_IMAGEWIDTH] === null 
+            || meta[Metadata.TIFF_IMAGEWIDTH] === undefined
+            || !Number.isFinite(meta[Metadata.TIFF_IMAGEWIDTH])){
+            console.log(`no width found for rendition of type ${meta[Metadata.DC_FORMAT]}`);
+            delete meta[Metadata.TIFF_IMAGEWIDTH];
+        }
+        if(meta[Metadata.TIFF_IMAGEHEIGHT] === null
```
tmathern, line 245: In case width is undefined, null or an Infinite (which is an invalid integer value in our case), delete the property to ensure no null value gets serialized/send.

lib/rendition.js
```diff
@@ -37,7 +37,7 @@ const Metadata = {
     DC_FORMAT: 'dc:format',
     ENCODING: 'repo:encoding',
     TIFF_IMAGEWIDTH: 'tiff:imageWidth',
-    TIFF_IMAGEHEIGHT: 'tiff:imageHeight'
+    TIFF_IMAGEHEIGHT: 'tiff:imageHeight' // should be TIFF_IMAGELENGTH: 'tiff:imageLength'
```
jdelbick, line 40: maybe make it a `// TODO`?

lib/rendition.js
```diff
@@ -236,6 +220,35 @@ class Rendition {
...
+            || meta[Metadata.TIFF_IMAGEWIDTH] === undefined
+            || !Number.isFinite(meta[Metadata.TIFF_IMAGEWIDTH])){
+            console.log(`no width found for rendition of type ${meta[Metadata.DC_FORMAT]}`);
+            delete meta[Metadata.TIFF_IMAGEWIDTH];
+        }
+        if(meta[Metadata.TIFF_IMAGEHEIGHT] === null 
+            || meta[Metadata.TIFF_IMAGEHEIGHT] === undefined
+            || !Number.isFinite(meta[Metadata.TIFF_IMAGEHEIGHT])){
+            console.log(`no height found for rendition of type ${meta[Metadata.DC_FORMAT]}`);
+            delete meta[Metadata.TIFF_IMAGEHEIGHT];
```
jdelbick, line 249: any unit test covering this?

jdelbick (approved):
LGTM, could you add some unit tests covering this?

lib/rendition.js
pheenomenon: If you mean AEM expects it as TIFF_IMAGELENGTH
```suggestion
    TIFF_IMAGEHEIGHT: 'tiff:imageHeight' // AEM expects it as TIFF_IMAGELENGTH: 'tiff:imageLength' 
```

pheenomenon (approved):
Wonder why the coverage increased even without additional unit tests.
Also the IT tests for smart-crop test is failing. looks unrelated to this change.

lib/rendition.js
pheenomenon: any idea why we call it TIFF_* ?

tmathern: We can't change this just now @pheenomenon , because this would pull in AEM testing (which I would like to avoid now, since in the end this is temporary until it's replaced by pipeline). AEM handles it properly already, we should just fix it eventually too.

tmathern:
@pheenomenon Test coverage can increse "artificially" if you add console.log lines, or the tests manage to go into more branches/statements because of code reorganization. Hence code coverage can increase without new tests, just by having more code accessible.

PR #141, opened 2021-08-18
CQ-4329896 - Added long parameter support with working test

badvision:
## Description

Shell script worker now takes parameters longer than 8000 characters and stores them in a temporary file.  The parameter value is replaced with the path of the temporary file and an environment variable FILE_PARAMS is set with the name(s) of parameters that had this substitution performed.

Fixes CQ-4329896 

## Types of changes

## Checklist:

- [ ] My change requires a change to the documentation. (no: this is a bugfix for a specific issue only)

### DIFF
--- a/lib/prepare.js
+++ b/lib/prepare.js
@@ -32,7 +32,7 @@ async function createDirectories() {
     //       <activationid>/   <- base
     //         in/             <- in
     //         out/            <- out
-    //         vars/           <- large variables, created as needed (if a variabel is over 128kb in size)
+    //         vars/           <- large variables, created as needed (if a variable is over 128kb in size)
 
     const directories = {};
 
--- a/lib/utils/utils.js
+++ b/lib/utils/utils.js
@@ -12,7 +12,6 @@
 
 'use strict';
 
-const tmp = require('tmp');
 const fs = require('fs-extra');
 
 // place for small utilities. larger parts should move into their own module.
@@ -67,16 +66,17 @@ function setConsoleLogPrefix(prefix) {
     };
 }
 
-// This should be roughly similar to ARG_MAX, some systems is 131072 per argument 
-// but we're going to be very conservative and cap the total size to be safe.
-const CMD_SIZE_LIMIT = 8000;
+// This should be roughly similar to ARG_MAX, some systems is 131072 bytes
+// but we're going to be very conservative and cap the total size much lower than that to be safe.
+const CMD_SIZE_LIMIT = 130500;
+const VAR_PATH = "./in/vars";
 
 /**
  * Only allows a specific number of bytes for variable names and values.  Any variable exceeding that will be stored in a file
  * and the variable value will contain the temp file location.
  * If this substitution occurs then an additional variable called FILE_PARAMS will contain the list of substituted variable names.
  * @param {Object} vars 
- * @param {Number} sizeLimit Size limit to uphold
+ * @param {Number} sizeLimit Size limit to uphold (number of bytes)
  * @returns Object of values from the provided one, modified as described.
  */
 function limitVariableSizes(vars, sizeLimit = CMD_SIZE_LIMIT) {
@@ -90,10 +90,10 @@ function limitVariableSizes(vars, sizeLimit = CMD_SIZE_LIMIT) {
         const newSize = totalSize + String(key).length + String(vars[key]).length;
         if (newSize >= sizeLimit) {
             // We have hit our limit and all variables must be stored in temp files now
-            const tmpFile = tmp.fileSync();
-            fs.writeSync(tmpFile.fd, vars[key]);
-            fs.closeSync(tmpFile.fd);
-            sizeLimitedVars[key] = tmpFile.name;
+            fs.mkdirpSync(VAR_PATH);
+            const tmpFile = `${VAR_PATH}/var-${key}-${(Math.random() + 1).toString(36).substring(2,8)}`;
+            fs.writeFileSync(tmpFile, vars[key]);
+            sizeLimitedVars[key] = tmpFile;
             // Now indicate that variable is a file path
             const fileParams = sizeLimitedVars.FILE_PARAMS || [];
             fileParams.push(key);
@@ -111,5 +111,6 @@ module.exports = {
     durationSec,
     timeUntilActivationTimeout,
     setConsoleLogPrefix,
-    limitVariableSizes
+    limitVariableSizes,
+    CMD_SIZE_LIMIT
 };
--- a/lib/utils/utils.js
+++ b/lib/utils/utils.js
@@ -13,7 +13,6 @@
 'use strict';
 
 const fs = require('fs-extra');
-const { spawnSync } = require('child_process');
 
 // place for small utilities. larger parts should move into their own module.
 
--- a/lib/utils/utils.js
+++ b/lib/utils/utils.js
@@ -87,6 +87,9 @@ function limitVariableSizes(vars, sizeLimit = CMD_SIZE_LIMIT) {
         if (size >= sizeLimit) {
             // We have hit our limit and all variables must be stored in temp files now
             fs.mkdirpSync(VAR_PATH);
+            // We want to salt the temp file name to avoid any chances of collision, however unlikely.
+            // Generate a random sequence of digits using [0-9A-Za-z] and pick 6 characters out of it.
+            // First two characters are always '1.' so we start with 2.
             const tmpFile = `${VAR_PATH}/var-${key}-${(Math.random() + 1).toString(36).substring(2,8)}`;
             fs.writeFileSync(tmpFile, vars[key]);
             sizeLimitedVars[key] = tmpFile;
--- a/lib/utils/utils.js
+++ b/lib/utils/utils.js
@@ -70,22 +70,8 @@ function setConsoleLogPrefix(prefix) {
 // This should be roughly similar to ARG_MAX, some systems is 131072 bytes
 // but we're going to be very conservative and cap the total size much lower than that to be safe.
 // POSIX suggestion is to go 2048 lower than the ARG_MAX value, as per https://www.in-ulm.de/~mascheck/various/argmax/
-const LEGACY_ARG_MAX = 131072;
-const CMD_SIZE_LIMIT = getArgMax();
+const CMD_SIZE_LIMIT = 131072;
 const VAR_PATH = "./in/vars";
-
-function getArgMax() {
-    try {
-        const ARG_MAX = parseInt(spawnSync("getconf", ["ARG_MAX"]).stdout.toString()) / 4;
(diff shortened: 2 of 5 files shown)


lib/utils/utils.js
```diff
@@ -64,8 +67,49 @@ function setConsoleLogPrefix(prefix) {
     };
 }
 
+// This should be roughly similar to ARG_MAX, some systems is 131072 per argument 
+// but we're going to be very conservative and cap the total size to be safe.
+const CMD_SIZE_LIMIT = 8000;
```
tmathern, line 72: Please precise the units - here I guess `bytes`.

lib/utils/utils.js
```diff
@@ -64,8 +67,49 @@ function setConsoleLogPrefix(prefix) {
     };
 }
 
+// This should be roughly similar to ARG_MAX, some systems is 131072 per argument
```
tmathern, line 70: Please add units explicitly (here it's bytes, right?).

lib/utils/utils.js
```diff
@@ -64,8 +67,49 @@ function setConsoleLogPrefix(prefix) {
...
+    );
+    const sizeLimitedVars = {};
+    let totalSize = 0;
+    for (const key of keys) {
+        const newSize = totalSize + String(key).length + String(vars[key]).length;
+        if (newSize >= sizeLimit) {
+            // We have hit our limit and all variables must be stored in temp files now
+            const tmpFile = tmp.fileSync();
+            fs.writeSync(tmpFile.fd, vars[key]);
+            fs.closeSync(tmpFile.fd);
```
tmathern, line 95: Idea: Maybe consider adding more smarts in the loop, to close the file descriptor only once after the loop, if it was opened/written to?

test/shellscript.test.js
```diff
@@ -496,5 +496,59 @@ describe("api.js (shell)", () => {
             assert.strictEqual(env.rendition_foobar, "Unicorn");
             assert.strictEqual(env.rendition_crop_x, "Unicorn");
         });
+
+        it("should pass variables larger than 8000 characters as a file", async () => {
```
tmathern, line 500: Also test the behavior when the length of parameters is exactly at the limit (aka 8000 chars).

lib/utils/utils.js
```diff
@@ -64,8 +67,49 @@ function setConsoleLogPrefix(prefix) {
...
+    const keys = Object.keys(vars).sort((a,b) => 
+        String(vars[a]).length - String(vars[b]).length
+    );
+    const sizeLimitedVars = {};
+    let totalSize = 0;
+    for (const key of keys) {
+        const newSize = totalSize + String(key).length + String(vars[key]).length;
+        if (newSize >= sizeLimit) {
+            // We have hit our limit and all variables must be stored in temp files now
+            const tmpFile = tmp.fileSync();
```
alexkli, line 93: As is, I assume the temp file will only be removed once the node process exits. But it should be removed after the current activation finishes. To achieve that, a temp file can be created inside the current activation folder - which is unique and will already be automatically removed after the processing.

lib/utils/utils.js
badvision: Well no I want to close the file and make sure there's no open descriptor.  The fact it's a temp file should be dealt with afterwards by the library.  And otherwise I don't want extra file descriptors laying around unnecessarily.  I get the temp file, write to it, close it and then record the path and move on.  I don't think there's any change needed here.

test/shellscript.test.js
badvision: I can appreciate why you'd want that but I think it's pretty sufficient, the goal is met by asserting that we cross well over the established 8000 character limit and the biggest variable is the only one which is written to a file.  The test itself creates an exactly 8000 character value and ensures that it is written exactly to a file.  If the test tries to make too many assumptions about how the code calculates the 8000 character threshold and how that's calculated exactly then it makes it more problematic to tune that behavior over time, I think.  Right now the calculation is based on variable name length + variable value length.   If we wanted to be really cautious then maybe adding two more for space delimiters and equal signs would be better?  What about quote marks?  Anyway, the point is I don't want the test to be overly prescriptive, I just want it to assert that at some point a line is crossed and that triggers the new logic.

As a side note, If we wanted to test more granularly then I think a better test is the utility function itself that does this work.  This would at least result in simpler looking tests that don't require actually executing a shell script.

lib/utils/utils.js
badvision: Oh!  That's right.  Good point, thank you.

lib/utils/utils.js
badvision: Anyway it seems that closing the file is not necessary if I use writeFileSync anyway...

badvision: Anyway it doesn't seem that closing the file is necessary when using writeFileSync...

test/shellscript.test.js
tmathern: Another think I was wondering about... How are we sure the test would actually fail? It could be running on machines with a larger limit... (in theory).

lib/utils/utils.js
tmathern: Then drop it alltogether?

badvision: Yes ma'am. :)

test/shellscript.test.js
badvision: The goal of the tests is not to generate the error condition, rather it is to assert the variables are written to files when we're expecting them to after they grow beyond a set size limit.  ARG_MAX and the resulting command buffer limitations really vary a lot by OS, so I have to assume lowest common denominator so we have consistent behavior.  I think 131072 is the right number, so to add some padding for extra characters, path names, etc I upped the limit to 130500.  That means we could have 572 other characters in the whole command that we're not counting before running into that issue again.  I think we can assume that's not an edge case of major concern.

lib/utils/utils.js
```diff
@@ -64,8 +66,50 @@ function setConsoleLogPrefix(prefix) {
     };
 }
 
+// This should be roughly similar to ARG_MAX, some systems is 131072 bytes
+// but we're going to be very conservative and cap the total size much lower than that to be safe.
+// POSIX suggestion is to go 2048 lower than the ARG_MAX value, as per https://www.in-ulm.de/~mascheck/various/argmax/
+const CMD_SIZE_LIMIT = 131072;
+const VAR_PATH = "./in/vars";
```
alexkli, line 73: Just need to ensure we don't add subfolders with name `vars` under `./in` otherwise. I don't think we do right now, but we should document. In case we e.g. unzip zip source files under `./in` and they contain a `var` folder.

See also https://github.com/adobe/asset-compute-sdk/blob/master/lib/prepare.js#L29-L34 (maybe good place to add a note about `vars` for now at least)

If we think there could be conflicts, we could have an `./var` or `./tmp` base folder (with e.g. `vars` inside, but open to more temp stuff) under the `<activationId>` folder.

badvision: Yes I agree we want to document this new behavior so folks know what to expect.  However, rather than expect people to unzip in the /in folder we should recommend against that.  A common directory is a lot like a global namespace, and therefore if they want to unzip into a folder without stepping on other files, they should create their own subfolder.

badvision: I moved the location to ./vars to prevent conflict and updated docs.

lib/prepare.js
```diff
@@ -32,6 +32,7 @@ async function createDirectories() {
     //       <activationid>/   <- base
     //         in/             <- in
     //         out/            <- out
+    //         vars/           <- large variables, created as needed (if a variabel is over 128kb in size)
```
tmathern, line 35: (typo)
```suggestion
    //         vars/           <- large variables, created as needed (if a variable is over 128kb in size)
```

lib/utils/utils.js
```diff
@@ -64,8 +66,46 @@ function setConsoleLogPrefix(prefix) {
...
+ * and the variable value will contain the temp file location.
+ * If this substitution occurs then an additional variable called FILE_PARAMS will contain the list of substituted variable names.
+ * @param {Object} vars 
+ * @param {Number} sizeLimit Size limit to uphold (number of bytes)
+ * @returns Object of values from the provided one, modified as described.
+ */
+function limitVariableSizes(vars, sizeLimit = CMD_SIZE_LIMIT) {
+    const sizeLimitedVars = {};
+    for (const key in vars) {
+        const size = String(key).length + String(vars[key]).length;
```
tmathern, line 86: Use string interpolation? (`...string...` <= the backquotes)

lib/utils/utils.js
```diff
@@ -64,8 +66,46 @@ function setConsoleLogPrefix(prefix) {
...
+        if (size >= sizeLimit) {
+            // We have hit our limit and all variables must be stored in temp files now
+            fs.mkdirpSync(VAR_PATH);
+            const tmpFile = `${VAR_PATH}/var-${key}-${(Math.random() + 1).toString(36).substring(2,8)}`;
```
tmathern, line 90: Why 36? Why 2,8? 
(My point: either use constants or add a comment where the values come from, because otherwise we'll forget).

test/shellscript.test.js
```diff
@@ -496,5 +497,66 @@ describe("api.js (shell)", () => {
...
+            createScript("worker.sh", `
+                env > envfile
+                echo ${testUtil.RENDITION_CONTENT} > $rendition
+            `);
+
+            // Build a random string that will exceed the command size limit.
+            let longStringValue = "";
+            // The random expression generates 5 each time.
+            for (let i = 0; i < CMD_SIZE_LIMIT; i += 5) {
+                longStringValue += (Math.random() + 1).toString(36).substring(2,7);
```
tmathern, line 511: Ditto

tmathern (approved):
Please add some details where some values come from -
"Conditional approval"

lib/utils/utils.js
badvision: Ok fair enough.

test/shellscript.test.js
badvision: Ok.

tmathern (approved):

lib/utils/utils.js
badvision: Both are valid stylistic choices, but I don't think it's worth re-triggering the build over it.

badvision: Also in this case using interpolation to avoid the concatenation operator actually makes it longer overall.  I have to safely cast to String to get the length, so the interpolated version would look like this (except convert single quotes to backticks):
'${String(key).length}${String(vars[key]).length}' -- Not sure it looks much better this way.

badvision: Since I want the length of a string, having javascript runtime parse a token pattern is less efficient than just calling the string function directly.  I don't think it sacrifices readability too much in this case.

PR #143, opened 2021-09-14
Always apply orientation for files that come to post processing

jdelbick:
## Description
While trying out a solution to use imagemagick for applying orientation, I came across these fixes required to use imagemagick (postprocessing) for applying orientation for non-pipeline cases.
- if a file comes to post processing because it needs orientation applied, make sure to send it through post processing
   - there was an edge case where a file would come to post processing and be the same source format, width and height and get rejected by post processing even though it needed orientation applied
   - the `needsOrientationApplied` function fixes this edge case by checking if the orientation value is >1 (1 or nothing means normal, no orientation needed to be applied)
- get rid of removing orientation metadata after applying `-auto-orient` (originally added here: https://github.com/adobe/asset-compute-sdk/pull/22/commits/9044bac0b142b7314585640fff894bf8072953cf)
    - old behavior:
       - we run the auto-orient cmd which applies orientation and sets the orientation exif value to 1 (`Horizontal (Normal)`)
       - then we remove the exif orientation data so it gets set to 0 (`Unknown`). 
    - new behaior:
       - we run the auto-orient cmd which applies orientation and sets the orientation exif value to 1 (`Horizontal (Normal)`)
- add test cases with a png source file that has orientation applied. these are more visual examples since you can visually see the orientation being applied

pre-req of this ticket: https://jira.corp.adobe.com/browse/NUI-1556

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:


lib/postprocessing/image.js
```diff
@@ -231,6 +231,12 @@ function hasJpegQuality(metadata, instructions) {
     return true;
 }
 
+function needsOrientationApplied(metadata) {
+    if (typeof(metadata.Orientation) === "number" && metadata.Orientation > 1) {
```
tmathern, line 235: A valid orientation will always be between 1 and 8. So I would test if the value is a Number and `>1 && <9`.

tmathern (changes requested):
I would make the test to know if re-orientation is needed or not more restrictive.

tmathern (approved):

lib/postprocessing/image.js
jdelbick: Done!

jdelbick:
fyi, dep mod tests are failing because `diff` fails for some post-processing tests since the orientation metadata changed on the rendition file: `Unknown (0)` -> `Horizontal (Normal)`
```
--------------------------------------------------------------------------------
Running validation: diff '/home/circleci/appLatest/worker-flite/action/tests/jpeg-with-embedded-icc-to-webp/rendition.webp' '/home/circleci/appLatest/worker-flite/action/.nui/2021_09_14T17_04_25_663Z/out/rendition0.webp'
Binary files /home/circleci/appLatest/worker-flite/action/tests/jpeg-with-embedded-icc-to-webp/rendition.webp and /home/circleci/appLatest/worker-flite/action/.nui/2021_09_14T17_04_25_663Z/out/rendition0.webp differ

!!! Validation failed: Rendition 'rendition0.webp' not as expected. Validate exit code was: 2
================================================================================
```
https://circle.ci.adobeitc.com/gh/nui/ci-opensource/20355/artifacts/0/test_logs/worker-flite/build/test.log
given that, is this considered a breaking change?

PR #166, opened 2022-02-16
Add `renditionCount` and `totalRenditionsSize` metrics to activation metrics

jdelbick:
## Description

As we start to work on optimization, we may find the need to look at historical data in new relic to make informed decisions. I noticed a few missing metrics that could help, so I am adding them now so we can start gathering data.

`renditionCount`: number of renditions attempted for processing
- this was only available in `process` metrics until now. If we look at action duration without knowing the amount of renditions, our decisions may have high variance. The number of renditions has a large impact on the action duration.

`totalRenditionsSize`: sum of the size of each rendition successfully produce
- I thought it would be interesting to track this. It is highly related to rendition count, but I see it as a helpful thing to look for, especially in workers that produce larger renditions like transcoding

The idea is if we have all the metrics we need in one event type, we can more easily analyze it. New Relic does not have a `join` clause to merge data between event types (ie no way to merge rendition and activation metric types on the same activation id).

Another interesting metric to add to activation metrics could be `renditionErrorCount`
### Testing
- added to unit test metrics check
- checked metrics from IT test integration: https://onenr.io/0BoQDZ084Ry
   - note: still need to duplicate the changes in pipeline

## Types of changes

## Checklist:

### DIFF
--- a/lib/worker.js
+++ b/lib/worker.js
@@ -68,7 +68,8 @@ class AssetComputeWorker {
         this.metrics.add({
             startWorkerDuration: Utils.durationSec(this.processingStartTime, this.workerStartTime),
             gatewayToProcessDuration: Utils.durationSec(this.params.times.gateway, this.params.times.process),
-            processToCoreDuration: Utils.durationSec(this.params.times.process, this.params.times.core)
+            processToCoreDuration: Utils.durationSec(this.params.times.process, this.params.times.core),
+            renditionCount: this.params.renditions && this.params.renditions.length
         });
 
         this.timers = {
@@ -78,6 +79,7 @@ class AssetComputeWorker {
             postProcessing:     new Timer(),
             upload:             new Timer()
         };
+        this.totalRenditionsSize = 0;
     }
 
     async compute(renditionCallback) {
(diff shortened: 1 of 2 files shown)


bobvanmanen (approved):
LGTM

pheenomenon (approved):

tmathern (approved):

lib/worker.js
```diff
@@ -392,6 +394,10 @@ class AssetComputeWorker {
 
         rendition.eventSent = true;
 
+        const renditionSize = rendition.size();
+        // track total rendition size to add to activation metrics
+        this.totalRenditionsSize += renditionSize;
```
tmathern, line 399: Only note: For large renditions, if there are a lot (a lot) of them, this can overflow.

jdelbick: true, I will create a safe add function that takes into account the max nodejs number: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER

fwiw, the max number is `9007199254740991`, which would translate to >9007 terabytes. I am pretty sure nui would timeout first if we were uploading that large amount of terabytes

jdelbick: buff-overflow

tmathern: Right, but adding the check won't hurt.

bobvanmanen: It's for metrics, i don't think we need to go too far. If we have renditions this large we can always improve.

jdelbick: Ok I will not add this check now, but we can always address it if we are seeing large amounts of renditions or very large renditions

PR #170, opened 2022-03-01
switch to only supporting node versions greater than 12, update deps

jdelbick:
## Description
I couldn't update a lot of the deps because the newest versions only support ESM:
- ex: https://github.com/TooTallNate/node-data-uri-to-buffer/releases/tag/4.0.0

## Types of changes

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,9 @@
     "type": "git",
     "url": "adobe/asset-compute-sdk"
   },
+  "engines": {
+    "node": ">=12.0.0"
+  },
   "dependencies": {
     "@adobe/asset-compute-commons": "^1.1.6",
     "@adobe/asset-compute-pipeline": "^7.1.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -10,6 +10,9 @@
     "type": "git",
     "url": "adobe/asset-compute-sdk"
   },
+  "engines": {
+    "node": ">=12.0.0"
+  },
```
jdelbick, line 15: this ensures it can only be installed with node >=12

tmathern: And we need this because there is a breakage on node 10? Or just because we don't support node 10 anymore?

jdelbick:
would this technically be a breaking release? wdyt? @tmathern @pheenomenon

pheenomenon:
As we would cause breakage for anyone on node 10,  I would go with a breaking change.

package.json
pheenomenon: I think because there is a breakage on node 10? with the dips updates.. , is that so @jdelbick ?

pheenomenon (approved):

PR #171, opened 2022-03-04
[ASSETS-6954] Use color profile file if available

tmathern:
## Description

Use the sRGB color profile file instead of default sRGB colorspace in post-processing when a default sRGB color profile file has been set (through an environment variable). Fallback to use default sRGB colorspace otherwise.

Fixes # [ASSETS-6954](https://jira.corp.adobe.com/browse/ASSETS-6954)

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -15,7 +15,7 @@ jobs:
     if: "!contains(github.event.head_commit.message, '[ci skip]')"
     strategy:
       matrix:
-        node-version: [12.16, 12.19, 14.16.1, 14.17]
+        node-version: [12.16, 12.19, 14.16.1, 14.17, 16.14]
 
     steps:
       - uses: actions/checkout@v2
--- a/lib/postprocessing/image.js
+++ b/lib/postprocessing/image.js
@@ -93,6 +93,8 @@ const EXTRA_INSTRUCTIONS = new Set([
     "watermark"
 ]);
 
+const SRGB_COLOR_PROFILE_FILE = process.env.SRGB_COLOR_PROFILE_FILE;
+
 function isSupportedOutputFormat(fmt) {
     return Object.keys(SUPPORTED_OUTPUT_FMT).includes(fmt);
 }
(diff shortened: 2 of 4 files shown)


.github/workflows/node.js.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     if: "!contains(github.event.head_commit.message, '[ci skip]')"
     strategy:
       matrix:
-        node-version: [12.16, 12.19, 14.16.1, 14.17]
+        node-version: [12.16, 12.19, 14.16.1, 14.17, 16.14]
```
tmathern, line 18: Adding node 16 to see if we're ready for next nodeJS LTS version...

lib/postprocessing/image.js
```diff
@@ -93,6 +93,8 @@ const EXTRA_INSTRUCTIONS = new Set([
     "watermark"
 ]);
 
+const SRGB_COLOR_PROFILE_FILE = process.env.SRGB_COLOR_PROFILE_FILE;
```
tmathern, line 96: If a colorprofile file should be used instead of default, the color profile path should be set in this env variable (eg. as env var in Docker if a docker image gets used).

lib/postprocessing/image.js
```diff
@@ -561,7 +563,14 @@ async function applyOutputProperties(img, instructions) {
     }
 
     // Most asset compute renditions are used in digital display context (web, screens), where RGB is required
-    img.colorspace('sRGB');
+    if(SRGB_COLOR_PROFILE_FILE){
+        console.log(`Using sRGB profile file: ${SRGB_COLOR_PROFILE_FILE}`);
+        img.profile(SRGB_COLOR_PROFILE_FILE);
+    } else {
+        console.log(`Using default imagemagick sRGB colorspace`);
+        img.colorspace('sRGB');
```
tmathern, line 571: Fallback - in case the file is not here (and for backwards compatibility).

jdelbick (commented):
could we add a `test-worker` test for this? or would that be too much effort since we would need to update the test container: https://github.com/adobe/asset-compute-sdk/blob/master/test-worker/manifest.yml#L7

tmathern:
> could we add a `test-worker` test for this? or would that be too much effort since we would need to update the test container: https://github.com/adobe/asset-compute-sdk/blob/master/test-worker/manifest.yml#L7

The tests are in the workers using a color profile. Here we won't set a color profile file for the tests, so I'd prefer the fallback ("old" behavior to use colorspace) to be used.

jdelbick (approved):
LGTM, please add a unit test for this

PR #172, opened 2022-03-09
[ASSETS-7826] Verify watermark object is not falsy

tmathern:
## Description

Verify the watermark object is not falsy, to avoid failing because of a watermark execution that should not be executed anyway. Regarding tests: Test would be in individual workers and ITs.

Fixes #[ASSETS-7826](https://jira.corp.adobe.com/browse/ASSETS-7826)

## Types of changes

## Checklist:

### DIFF
--- a/lib/postprocessing/image.js
+++ b/lib/postprocessing/image.js
@@ -440,7 +440,7 @@ async function applyWatemark(img, intermediateRendition, rendition, directories)
     const instructions = rendition.instructions;
     const targetBoundingBox = new Dimensions(instructions.width, instructions.height);
 
-    if(!instructions.watermark){
+    if (!instructions.watermark) {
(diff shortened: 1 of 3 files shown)


lib/postprocessing/image.js
```diff
@@ -440,6 +440,11 @@ async function applyWatemark(img, intermediateRendition, rendition, directories)
     const instructions = rendition.instructions;
     const targetBoundingBox = new Dimensions(instructions.width, instructions.height);
 
+    if (!instructions.watermark) {
+        console.log(`Cannot apply watermark: watermark object is invalid (falsy)`);
```
tmathern, line 444: I chose not to throw an exception, because a falsy value could be "legitimate".

tmathern: Note: `instructions.watermark === 'true'` should fail asset processing.

PR #174, opened 2022-03-24
ASSETS-8775 add predicted run duration to activation metrics

jdelbick:
## Description

jira ticket: https://jira.corp.adobe.com/browse/ASSETS-8775
subtask: https://jira.corp.adobe.com/browse/ASSETS-9897
to enable fast feedback loop of predicted run duration vs actual run duration
## Types of changes

## Checklist:

### DIFF
--- a/lib/webaction.js
+++ b/lib/webaction.js
@@ -73,12 +73,17 @@ function getParams(params) {
     const requestId = headers["x-request-id"];
     const auth = getAuth(headers);
 
+    if (params.predictedRunDuration) {
+        console.log('Predicted run duration for the activation', params.predictedRunDuration);
+    }
+
     params.metrics.add({
         clientId: auth.clientId,
         appName: auth.appName,
         orgId: auth.orgId,
         orgName: auth.orgName,
-        requestId: requestId
+        requestId: requestId,
+        predictedRunDuration: params.predictedRunDuration
     });
 
     return {
--- a/lib/webaction.js
+++ b/lib/webaction.js
@@ -117,10 +117,9 @@ function webaction(main) {
     return actionWrapper(async (params={}) => {
         // Asset Compute Throttling Service metrics
         // for quick feedback loop, store predicted run duration next to actual run duration
-        if (params && params.metrics && params.predictedRunDuration) {
-            console.log('params.metrics and params.predictedRunDuration are available');
+        if (params && params.metrics && params.estimatedRunDuration) {
             params.metrics.add({
-                predictedRunDuration: params.predictedRunDuration
+                estimatedRunDuration: params.estimatedRunDuration
             });
         }
         // if web action (custom worker), then invoke itself asynchronously
--- a/lib/webaction.js
+++ b/lib/webaction.js
@@ -90,8 +90,7 @@ function getParams(params) {
         newRelicEventsURL: params.newRelicEventsURL,
         newRelicApiKey: params.newRelicApiKey,
         times: params.times,
-        customWorker: true,
-        predictedRunDuration: params.predictedRunDuration
+        customWorker: true
     };
 }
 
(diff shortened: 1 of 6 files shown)


lib/webaction.js
```diff
@@ -90,7 +90,8 @@ function getParams(params) {
         newRelicEventsURL: params.newRelicEventsURL,
         newRelicApiKey: params.newRelicApiKey,
         times: params.times,
-        customWorker: true
+        customWorker: true,
+        predictedRunDuration: params.predictedRunDuration
```
jdelbick, line 94: do we want/need predicted run duration in custom workers? IMO we don't need it because we will never be able to predict or control a custom worker time length/cost as it can change without us knowing.

also, just realized, we don't have a cost value/duration for custom workers in core cost values. I think we should have static values for these for all custom workers. this will be inaccurate of course since custom workers can choose their own memory/timeout values but we need to have something thats onthe higher end. wdyt

lib/webaction.js
```diff
@@ -116,6 +116,14 @@ async function invokeSelfAsync(params) {
...
     // actionWrapper -> webaction -> main (worker)
     return actionWrapper(async (params={}) => {
+        // Asset Compute Throttling Service metrics
+        // for quick feedback loop, store predicted run duration next to actual run duration
+        if (params && params.metrics && params.predictedRunDuration) {
+            console.log('params.metrics and params.predictedRunDuration are available');
+            params.metrics.add({
+                predictedRunDuration: params.predictedRunDuration
+            });
+        }
```
jdelbick, line 126: putting it here makes it account for all the different cases:
- pipeline, non pipeline
- shellScriptWorker
- batchWorker
- regular Worker
and it works because actionWrapper adds metrics here: https://github.com/adobe/asset-compute-commons/blob/master/lib/wrap.js#L50

lib/webaction.js
```diff
@@ -116,6 +115,14 @@ async function invokeSelfAsync(params) {
 function webaction(main) {
     // actionWrapper -> webaction -> main (worker)
     return actionWrapper(async (params={}) => {
+        // Asset Compute Throttling Service metrics
+        // for quick feedback loop, store predicted run duration next to actual run duration
+        if (params && params.metrics && params.predictedRunDuration) {
+            console.log('params.metrics and params.predictedRunDuration are available');
+            params.metrics.add({
+                predictedRunDuration: params.predictedRunDuration
```
jdelbick, line 123: qs, we use `estimatedRunDuration` for costTable metrics but `predictedRunDuration` here. do we want `estimatedRunDuration` to be 1:1? (this will require a small change in core mob-session)

pheenomenon: its the same as calling `estimatedRunDuration` here correct ? +1 on keeping the attribute name same at both places.

lib/webaction.js
```diff
@@ -116,6 +115,14 @@ async function invokeSelfAsync(params) {
 function webaction(main) {
     // actionWrapper -> webaction -> main (worker)
     return actionWrapper(async (params={}) => {
+        // Asset Compute Throttling Service metrics
+        // for quick feedback loop, store predicted run duration next to actual run duration
+        if (params && params.metrics && params.predictedRunDuration) {
+            console.log('params.metrics and params.predictedRunDuration are available');
```
tmathern, line 121: Drop this log, it's not that useful for debugging.

lib/webaction.js
tmathern: +1 for same name

tmathern (approved):

lib/webaction.js
pheenomenon: Will this code will run for actions like `core`, `api-register` and I think also for custom-worker.  Which may be ok, as they won't have `params.predictedRunDuration`

pheenomenon (approved):

adamcin (approved):

lib/webaction.js
jdelbick: ACK, will make the change

jdelbick: just changed in core also: https://git.corp.adobe.com/nui/core/pull/326/commits/9dd71e5a8faf28d63f3ae23e380e636499a2d882

lib/webaction.js
jdelbick: this code won't run for core and api-register and core because they use the actionWrapper directly from commons: https://git.corp.adobe.com/nui/core/blob/master/core.js#L22
but yeah the code is designed so if there is no predictedRunDuration or params.metrics, it just will ignore it

PR #182, opened 2022-05-25
ASSETS-8997 add support for serviceoverload rendition_failed reason

adamcin:
jira: https://jira.corp.adobe.com/browse/ASSETS-8997
depends on upstream [@adobe/asset-compute-commons PR#71](https://github.com/adobe/asset-compute-commons/pull/71)

Adds support for `ServiceOverLoad` rendition_failed reason in SDK in order to prevent coercion of `"reason": "ServiceOverLoad"` to GenericError prior to sending the event to Adobe I/O.

### DIFF
--- a/index.js
+++ b/index.js
@@ -21,7 +21,8 @@ const {
     SourceFormatUnsupportedError,
     SourceUnsupportedError,
     SourceCorruptError,
-    RenditionTooLarge
+    RenditionTooLarge,
+    ServiceOverLoadError
 } = require('@adobe/asset-compute-commons');
 
 // -----------------------< exports >-----------------------------------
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -21,7 +21,8 @@ const {
     SourceFormatUnsupportedError,
     SourceUnsupportedError,
     SourceCorruptError,
-    RenditionTooLarge
+    RenditionTooLarge,
+    ServiceOverLoadError
```
tmathern, line 25: See note from https://github.com/adobe/asset-compute-commons/pull/71, which may lead to changes here.

pheenomenon (approved):

PR #214, opened 2024-01-10
ASSETS-23937 - Support setting a color as background color

tmathern:
## Description

Fixes # (issue)
- Add support for the instruction that is already used in some workers to set a background color, `pdfbgcolor`;
- Allowed color schemes: hex color or valid numerical value for a color channel (the same value will be used for the three RGB channels);
- Does not do any aio/asset-compute-plugin dependencies update for now, as something seems to be broken when running tests (see https://github.com/adobe/asset-compute-sdk/pull/213).

## Types of changes

## Checklist:


.github/workflows/node.js.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     if: "!contains(github.event.head_commit.message, '[ci skip]')"
     strategy:
       matrix:
-        node-version: [14.16.1, 14.17]
+        node-version: [14.17]
```
tmathern, line 18: Dropping the oldest version.
I didn't update to a newer one because currently it breaks the test pipeline, since dependency hell in https://github.com/adobe/asset-compute-sdk/pull/213.

test-worker/test/asset-compute/worker/png-background-color/params.json
```diff
@@ -0,0 +1,4 @@
+{
+    "fmt": "png",
+    "pdfbgcolor": "#0EF"
```
tmathern, line 3: This tests mimics adding a color as a background to any image with a transparent background

sudheendrakatikar (approved):

lib/postprocessing/image.js
```diff
@@ -639,22 +637,55 @@ function convertSvg(img, instructions) {
...
+        return `rgb(${inputBackgroundColor},${inputBackgroundColor},${inputBackgroundColor})`;
     } else {
-        img.background(DEFAULT_BACKGROUND_COLOR); // Change tranparency background to default background color 
-        img.flatten();  // forces the image on its background to "commit" the background
+        throw new GenericError(`Invalid background color values set for RGB channels in instructions: ${instructions.pdfbgcolor}`);
     }
+}
 
+function handleTransparency(img, instructions) {
+    if(instructions.pdfbgcolor) {
```
pheenomenon, line 675: if we expect the supported format to be `png` then the condition should specify it. 
if(instructions.pdfbgcolor && instructions.fmt === 'png') {

tmathern: at that point we'd support anything, since the whole imagemagick command will do the conversion. It's the input we expect to be PNG.

tmathern:
(Failing dependent tests expected, since the failing worker for PDF needs some adaptation to avoid double-processing)

PR #224, opened 2026-05-27
ASSETS-68731: add onAfterRendition finalizer hook in postProcess

pheenomenon:
for [ASSETS-68731](https://jira.corp.adobe.com/browse/ASSETS-68731) and ASSETS-67356 
Generic hook that runs registered callbacks against the final rendition after SDK image post-processing (or in its place when post-processing is skipped). Enables workers to defer manifest/metadata operations until after imagemagick re-encode, which would otherwise strip them.

Each finalizer is wrapped in its own try/catch so a failure cannot break the upload path nor prevent other finalizers from running. Defensive against non-array values on _postProcessFinalizers.

Hook is opaque to the SDK — no c2pa or other domain knowledge here. Workers register finalizers by pushing onto rendition._postProcessFinalizers (documented contract). The Rendition class itself lives in @adobe/asset-compute-pipeline and is intentionally not modified.

## Description

Fixes # (issue)

## Types of changes

## Checklist:


dconnelly84 (approved):

lib/worker.js
```diff
@@ -339,6 +350,18 @@ class AssetComputeWorker {
...
 
+            // Run onAfterRendition finalizers against the final rendition,
+            // whether or not SDK image post-processing actually transformed it.
+            // Each finalizer is wrapped so one failure cannot break upload nor
+            // prevent subsequent finalizers from running.
+            for (const fn of finalizers) {
+                try {
+                    await fn(rendition);
+                } catch (err) {
+                    console.log(`onAfterRendition finalizer failed (non-fatal): ${err && err.message || err}`);
```
sashoscoder, line 361: so, in this case you're just consuming the error. If anything fails in the finalizer, then it's consumed and a wrong rendition would be written to the upload path right? instead of failing the rendition?

pheenomenon: yes, I wanted it to be really defensive, and not cause a regression. we can revisit this to throw errors later, but in this case for c2pa manifest, even if the manifest application fails, we do not fail the rendition.

sashoscoder (approved):

